PR #25, opened 2014-07-19
Support Yields section.

chebee7i:
For generators, a Yields section is more helpful than a Returns section.

For some reason, I thought Yields was already supported by numpydoc. I guess not.
Napoleon supports it: http://sphinxcontrib-napoleon.readthedocs.org/

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -292,8 +292,9 @@ class NumpyDocString(object):
         for (section,content) in self._read_sections():
             if not section.startswith('..'):
                 section = ' '.join([s.capitalize() for s in section.split(' ')])
-            if section in ('Parameters', 'Returns', 'Raises', 'Warns',
-                           'Other Parameters', 'Attributes', 'Methods'):
+            if section in ('Parameters', 'Returns', 'Yields', 'Raises',
+                           'Warns', 'Other Parameters', 'Attributes',
+                           'Methods'):
                 self[section] = self._parse_param_list(content)
             elif section.startswith('.. index::'):
                 self['index'] = self._parse_index(section, content)
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -289,7 +289,17 @@ class NumpyDocString(object):
         self._doc.reset()
         self._parse_summary()
 
-        for (section,content) in self._read_sections():
+        sections = list(self._read_sections())
+        section_names = set([section for section, content in sections])
+
+        has_returns = 'Returns' in section_names
+        has_yields = 'Yields' in section_names
+        # We could do more tests, but we are not. Arbitrarily.
+        if has_returns and has_yields:
+            msg = 'Docstring contains both a Returns and Yields section.'
+            raise ValueError(msg)
+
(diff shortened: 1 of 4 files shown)


rgommers:
Here: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

There are separate sections for documenting classes, modules, etc., so no reason not to have one for generators.

patricksnape:
I would like to +1 this functionality.

numpydoc/docscrape.py
```diff
@@ -291,8 +292,9 @@ def _parse(self):
         for (section,content) in self._read_sections():
             if not section.startswith('..'):
                 section = ' '.join([s.capitalize() for s in section.split(' ')])
-            if section in ('Parameters', 'Returns', 'Raises', 'Warns',
-                           'Other Parameters', 'Attributes', 'Methods'):
+            if section in ('Parameters', 'Returns', 'Yields', 'Raises',
+                           'Warns', 'Other Parameters', 'Attributes',
+                           'Methods'):
```
tacaswell: There should probably be some logic here to check that both Returns and Yields are not specified.

numpydoc/docscrape_sphinx.py
```diff
@@ -63,6 +63,23 @@ def _str_returns(self):
                 out += ['']
         return out
 
+    def _str_yields(self):
```
tacaswell: This is a copy-paste of `_str_param_list` with `'Yields'` hard-coded?

numpydoc/tests/test_docscrape.py
```diff
@@ -164,6 +164,29 @@ def test_returns():
...
+------
+a : int
+    The number of apples.
+b : int
+    The number of bananas.
+
+"""
+    doc = NumpyDocString(doc_text)
+    section = doc['Yields']
+    print(section)
```
pv: Shouldn't have prints here

tacaswell:
The `test_str` and `test_sphinx_str` test functions should also be expanded to verify that the properly output the `Yield` section as well.

numpydoc/docscrape.py
chebee7i: I can add that, but note, it doesn't look like we verify other stuff...like having repeated sections or missing a returns section.

tacaswell: Fair enough.

numpydoc/docscrape_sphinx.py
```diff
@@ -225,6 +225,7 @@ def __str__(self, indent=0, func_role="obj"):
         out += self._str_extended_summary()
         out += self._str_param_list('Parameters')
         out += self._str_returns()
```
tacaswell: I know it is not required, but for readability it might be worth passing `'Returns'` here.  This is really minor, ignore it if you want.

chebee7i:
Any update on this?

charris:
Any progress here? There is a numpy documentation PR that I have waiting on this.

stefanv:
Looks good to me.

pv:
Thanks, merged

PR #39, opened 2015-09-23
MAINT inspect.getfullargspec is deprecated in Py3.5

ogrisel:
And the code is much nicer with the new signature API.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -460,15 +460,17 @@ class FunctionDoc(NumpyDocString):
         if not self['Signature'] and func is not None:
             func, func_name = self.get_func()
             try:
-                # try to read signature
-                if sys.version_info[0] >= 3:
-                    argspec = inspect.getfullargspec(func)
+                if hasattr(inspect, 'signature'):
+                    signature = str(inspect.signature(func))
                 else:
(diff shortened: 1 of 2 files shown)


numpydoc/docscrape.py
```diff
@@ -460,15 +460,18 @@ def __init__(self, func, role='func', doc=None, config={}):
...
-                argspec = argspec.replace('*', '\*')
-                signature = '%s%s' % (func_name, argspec)
-            except TypeError as e:
+                    # try to read signature, backward compat for older Python
+                    if sys.version_info[0] >= 3:
+                        argspec = inspect.getfullargspec(func)
+                    else:
+                        argspec = inspect.getargspec(func)
+                    argspec = inspect.formatargspec(*argspec)
+                    argspec = argspec.replace('*', '\*')
```
pv: Does inspect.signature deal with this?

ogrisel: hum, probably not indeed. This is for rst escaping right?

ogrisel: I amended my commit to fix this and added a check in the tests.

ogrisel:
The tests pass including the newly introduced check on the escaping of the star character.

PR #40, opened 2015-10-05
better error messages from NumpyDocString

larsmans:
`make html` wouldn't say where the error originates. This way, we get at least a hint.


stefanv:
Looks fine to me.

PR #42, opened 2015-11-17
numpydoc.docscrape: Restore support for Python 3.4.

androomerrill:
Fixing backwards compatibility bug introduced in commit
254494c, which broke support for Python 3.4 when passing
the builtin method 'mro' of 'type' objects to
'inspect.signature', raising a 'ValueError'.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -473,9 +473,9 @@ class FunctionDoc(NumpyDocString):
         if not self['Signature'] and func is not None:
             func, func_name = self.get_func()
             try:
-                if hasattr(inspect, 'signature'):
+                try:
                     signature = str(inspect.signature(func))
-                else:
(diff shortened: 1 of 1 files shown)


rgommers:
@androomerrill I'll test this with Python 3.4. But can you explain how 254494c broke something? I don't see anything in there except PEP8 style cleanups.

androomerrill:
@rgommers, sorry about that. I'm not sure where I pulled that commit hash from, but it's clearly wrong. I'll fix it asap.

rgommers:
I can build the numpy docs with Python 3.4 just fine.

rgommers:
ok seeing the issue now.

rgommers:
The offending commit was b9c4c90d

rgommers:
This fixes the issue, so in it goes. Thanks @androomerrill

PR #46, opened 2015-11-25
Set parallel_write_safe in setup metadata

jschueller:


jschueller:
@pv, hi, is this ok ?

pv:
I don't know: is the extension parallel write safe?

jschueller:
@pv, not sure, there's some file io (eventually not be parallel-safe) that only occur in the deprecated plot_directive module, so I'd say it must be ok.

stefanv:
We had to modify the scikit-image build scripts to run Sphinx single-threaded, otherwise references get messed up.  Does that impact on this PR?

rgommers:
``parallel_write_safe`` was only added in Sphinx 1.3, and we support 1.2.3, so not sure this is okay. More importantly, the default value is True (http://www.sphinx-doc.org/en/1.5.1/extdev/#extension-metadata), so this PR is unnecessary. Given that there's also a merge conflict, I'll close it.

Thanks @jschueller, all

PR #48, opened 2016-01-31
REL: use the distutils sdist command instead of the setuptools one.

rgommers:
This removed unwanted useless files: an empty ``setup.cfg`` and a ``numpy.egg-info`` directory.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,8 @@
+include MANIFEST.in
 recursive-include numpydoc/tests *.py
 include *.txt
+include *.rst
+
+# Exclude what we don't want to include
+prune */__pycache__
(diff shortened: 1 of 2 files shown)


rgommers:
I'll merge this in a day or two if there are no comments by then, do some final testing and tag ``v0.6``.

rgommers:
OK merged. I'll build a few docs from an installed sdist as a final test.

rgommers:
For the final release commit I also want to change the version number to 0.6.0 (instead of 0.6), add some PyPi classifiers for supported Python versions and bump the status from Alpha to Beta. See https://github.com/rgommers/numpydoc/commit/ef988a4a4658c991f4445f6241ab02d74710c6e3

PR #51, opened 2016-03-02
Make estension version available from sphinx

callegar:
This is meant to allow other extensions to depend on a specific version of numpydoc or to base their behavior on the version of numpydoc that is available.


rgommers:
This is pretty straightforward, so can review/merge before the 0.7.0 release. @callegar sorry no one got around to reviewing it earlier. Can you rebase?

jschueller:
there's already a version string in __init__.py

callegar:
Yes, it has been added 6 months ago.

Now, a lot of time has passed, and I do not remember 100% of the matter, but if I do remember properly, it does not seem to solve the whole of the problem.

- It cannot be easily read from setup.py, so this information needs to be duplicated there and duplication is errorprone
- It is hardwired in __init__, and thus harder to automatically generate based on VCS tags, if in the future there is the intention to do so (e.g. by python-versioneer)
- It is still not exported in the package metadata as sphinx expects.

rgommers:
Still needs a rebase. ``_version.py`` is good to keep, the regular expression version parsing needs to go (current ``setup.py`` has a much more readable way of parsing the version string).

Looks good to merge once that's done.

rgommers:
Sorry, just see that I missed gh-94.

PR #53, opened 2016-04-06
allow extending NumpyDocString sections

haobibo:
Current implementation of NumpyDocString define sections in
`self._parsed_data` in `NumpyDocString.__init__()`, which is not
convenient for developers to extend this class to add more sections. By
setting section definitions as a class static variable, it will be more
convenient for developers to extend more sections by changing the class
static variable `sections`.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -97,30 +97,32 @@ class ParseError(Exception):
 
 
 class NumpyDocString(collections.Mapping):
+    sections = {
+        'Signature': '',
+        'Summary': [''],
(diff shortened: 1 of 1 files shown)


stefanv:
I guess the purpose of a documentation standard is, in part, not to be extended.  That said, this change is innocuous enough, so I'm ok with it.

pv:
Probably should be `self._parsed_data = copy.deepcopy(self.sections)` 
because it contains mutable lists.

haobibo:
@pv already changed to deep copy.

Issue #55, opened 2016-06-26
Use 'install_requires' instead of 'requires' in setup.py


dukebody:
I've noticed that sphinx is not installed automatically when the package is installed via pip. Actually it doesn't either when doing `python setup.py develop`.

I believe this is because the wrong keyword `requires` is being used in [`setup.py`](https://github.com/numpy/numpydoc/blob/master/setup.py), when it should be `install_requires`: http://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=setup#new-and-changed-setup-keywords  I've read that setuptools monkeypatches distutils, so even if in that file the `setup` function comes from distutils, `install_requires` should work.

Changing the keyword effectively triggers installation of sphinx when installing numpydoc. PR will come shortly.

rgommers:
fixed in gh-56

PR #56, opened 2016-06-26
Use setuptools 'install_requires' instead of 'requires' in setup.py.

dukebody:
This will trigger sphinx installation when installing numpydoc (tested locally). See https://github.com/numpy/numpydoc/issues/55


rgommers:
merged, thanks @dukebody

PR #61, opened 2016-09-07
add (possibly slightly off) source lines to mangled docstrings.

### LINKED ISSUE #52: Errors result in huge, verbose output
It seems when there's an error in sphinx, you get a printout of the following form
```
path/to/filename.rst:77: ERROR: Unknown target name: "children".
```

When an error occurs in a numpydoc-ed class, the filename part is replaced with a representation of a python list of all the lines (in rst-form, post numpydoc mangling) which can be quite long. It fills up my bash history and makes it incredibly hard to see the actual error message.

```
['n_components_  (int) The estimated number of connected components in the graph.  ', '    children_  (array-like, shape (n_nodes-1, 2)) The children of each non-leaf node. Values less than `n_samples` correspond to leaves of the tree which are the original samples. A node `i` greater than or equal to `n_samples` is a non-leaf node and has children `children_[i - n_samples]`. Alternatively at the i-th iteration, children[i][0] and children[i][1] are merged to form node `n_samples + i`  ', '=============  ==========', '', '.. rubric:: Methods', '', '.. autosummary::', '', '   fit', '   fit_predict', '   fit_transform', '   get_params', '   partial_predict', '   partial_transform', '   predict', '   set_params', '   summarize', '   transform', '', '.. automethod:: __init__', '', '', '.. rubric:: Methods', '', '.. autosummary::', '', '   ~Ward.__init__', '   ~Ward.fit', '   ~Ward.fit_predict', '   ~Ward.fit_transform', '   ~Ward.get_params', '   ~Ward.partial_predict', '   ~Ward.partial_transform', '   ~Ward.predict', '   ~Ward.set_params', '   ~Ward.summarize', '   ~Ward.transform', '', '.. py:method:: Ward.fit(sequences, y=None)', '   :module: msmbuilder.cluster', '', '   ', '   ', '   Fit the  clustering on the data', '   ', '   ', '   :Parameters:', '   ', '       **sequences** : list of array-like, each of shape [sequence_length, n_features]', '   ', '           A list of multivariate timeseries. Each sequence may have', '           a different length, but they all must have the same number', '           of features.', '   ', '   :Returns:', '   ', '       self', '   ', '           ', '   ', '', '.. py:method:: Ward.fit_predict(sequences, y=None)', '   :module: msmbuilder.cluster', '', '   ', '   ', '   Performs clustering on X and returns cluster labels.', '   ', '   ', '   :Parameters:', '   ', '       **sequences** : list of array-like, each of shape [sequence_length, n_features]', '   ', '           A list of multivariate timeseries. Each sequence may have', '           a different length, but they all must have the same number', '           of features.', '   ', '   :Returns:', '   ', '       **Y** : list of ndarray, each of shape [sequence_length, ]', '   ', '           Cluster labels', '   ', '', '.. py:method:: Ward.fit_transform(sequences, y=None)', '   :module: msmbuilder.cluster', '', '   ', '   ', '   Alias for fit_predict', '   ', '   ', '', '.. py:method:: Ward.get_params(deep=True)', '   :module: msmbuilder.cluster', '', '   ', '   ', '   Get parameters for this estimator.', '   ', '   ', '   :Parameters:', '   ', '       **deep: boolean, optional**', '   ', '           If True, will return the parameters for this estimator and', '           contained subobjects that are estimators.', '   ', '   :Returns:', '   ', '       **params** : mapping of string to any', '   ', '           Parameter names mapped to their values.', '   ', '', '.. py:method:: Ward.partial_predict(X, y=None)', '   :module: msmbuilder.cluster', '', '   ', '   ', '   Predict the closest cluster each sample in X belongs to.', '   ', '   In the vector quantization literature, `cluster_centers_` is called', '   the code book and each value returned by `predict` is the index of', '   the closest code in the code book.', '   ', '   :Parameters:', '   ', '       **X** : array-like shape=(n_samples, n_features)', '   ', '           A single timeseries.', '   ', '   :Returns:', '   ', '       **Y** : array, shape=(n_samples,)', '   ', '           Index of the cluster that each sample belongs to', '   ', '', '.. py:method:: Ward.partial_transform(X)', '   :module: msmbuilder.cluster', '', '   ']:77: ERROR: Unknown target name: "children".
```

Can anyone point me to where I might investigate where this filename is getting (re-)set?

amueller:
Fixes #52 to some degree.
This adds references to the original source files instead of gigantic output.
However, the numpy mangling might add lines and so the numbers might be slightly off.
I would argue this is already a big improvement though.

---
This change is [Reviewable](https://reviewable.io/reviews/numpy/numpydoc/61)

### DIFF
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -29,7 +29,6 @@ if sphinx.__version__ < '1.0.1':
     raise RuntimeError("Sphinx 1.0.1 or newer is required")
 
 from .docscrape_sphinx import get_doc_object, SphinxDocString
-from sphinx.util.compat import Directive
 
 if sys.version_info[0] >= 3:
     sixu = lambda s: s
@@ -139,7 +138,7 @@ def setup(app, get_doc_object_=get_doc_object):
     # Extra mangling domains
     app.add_domain(NumpyPythonDomain)
     app.add_domain(NumpyCDomain)
-    
+
     metadata = {'parallel_read_safe': True}
     return metadata
 
@@ -190,6 +189,62 @@ class NumpyCDomain(ManglingDomainBase, CDomain):
     }
 
 
+def match_items(lines, content_old):
+    """Create items for mangled lines.
+
+    This function tries to match the lines in ``lines`` with the items (source
+    file references and line numbers) in ``content_old``. The
+    ``mangle_docstrings`` function changes the actual docstrings, but doesn't
+    keep track of where each line came from. The manging does many operations
+    on the original lines, which are hard to track afterwards.
+
+    Many of the line changes come from deleting or inserting blank lines. This
+    function tries to match lines by ignoring blank lines. All other changes
+    (such as inserting figures or changes in the references) are completely
+    ignored, so the generated line numbers will be off if ``mangle_docstrings``
+    does anything non-trivial.
+
+    This is a best-effort function and the real fix would be to make
+    ``mangle_docstrings`` actually keep track of the ``items`` together with
+    the ``lines``.
+
+    Examples
+    --------
+    >>> lines = ['', 'A', '', 'B', '   ', '', 'C', 'D']
+    >>> lines_old = ['a', '', '', 'b', '', 'c']
+    >>> items_old = [('file1.py', 0), ('file1.py', 1), ('file1.py', 2),
+    ...              ('file2.py', 0), ('file2.py', 1), ('file2.py', 2)]
+    >>> content_old = ViewList(lines_old, items=items_old)
+    >>> match_items(lines, content_old) # doctest: +NORMALIZE_WHITESPACE
+    [('file1.py', 0), ('file1.py', 0), ('file2.py', 0), ('file2.py', 0),
+     ('file2.py', 2), ('file2.py', 2), ('file2.py', 2), ('file2.py', 2)]
+    >>> # first 2 ``lines`` are matched to 'a', second 2 to 'b', rest to 'c'
+    >>> # actual content is completely ignored.
+
+    Notes
+    -----
+    The algorithm tries to match any line in ``lines`` with one in
+    ``lines_old``.  It skips over all empty lines in ``lines_old`` and assigns
+    this line number to all lines in ``lines``, unless a non-empty line is
+    found in ``lines`` in which case it goes to the next line in ``lines_old``.
+
+    """
+    items_new = []
+    lines_old = content_old.data
+    items_old = content_old.items
+    j = 0
+    for i, line in enumerate(lines):
+        # go to next non-empty line in old:
+        # line.strip() checks whether the string is all whitespace
+        while j < len(lines_old) - 1 and not lines_old[j].strip():
+            j += 1
+        items_new.append(items_old[j])
+        if line.strip() and j < len(lines_old) - 1:
+            j += 1
+    assert(len(items_new) == len(lines))
+    return items_new
+
+
 def wrap_mangling_directive(base_directive, objtype):
     class directive(base_directive):
         def run(self):
@@ -205,7 +260,10 @@ def wrap_mangling_directive(base_directive, objtype):
 
             lines = list(self.content)
             mangle_docstrings(env.app, objtype, name, None, None, lines)
-            self.content = ViewList(lines, self.content.parent)
+            if self.content:
+                items = match_items(lines, self.content)
+                self.content = ViewList(lines, items=items,
+                                        parent=self.content.parent)
 
             return base_directive.run(self)
 


pv:
Thanks! How does the new error message look after this?

amueller:
same as without numpydoc:
``/home/andy/checkout/sphinx_autodoc_debug/main.py:docstring of main.my_function:20: ERROR: Unknown target name: "reference".``
(using this minimal example: https://github.com/amueller/sphinx_autodoc_debug)

amueller:
It would be great to have the line number in the ``.py`` file but that is an issue of how autodoc (I think) formats the source. Now at least numpydoc is not messing up the stuff that autodoc generates.

amueller:
I found out that there were just a couple of empty lines added and other empty lines removed in the parsing which made the line length different . I'm trying to figure that out, so we can get better matching line numbers.

amueller:
ok I did a rough heuristic for better lines. The main reason I prefer the heuristic is that classes are single documents, and I want to know which method the problem is in. This information gets lost otherwise. My heuristic might be off, but I think it's a net improvement.

I ran it over the scikit-learn docs and it didn't break in a horrible way.

numpydoc/numpydoc.py
```diff
@@ -183,6 +183,31 @@ class NumpyCDomain(ManglingDomainBase, CDomain):
...
+    are always empty.
+    """
+    items_new = []
+    lines_old = content_old.data
+    items_old = content_old.items
+    j = 0
+    for i, line in enumerate(lines):
+        # go to next non-empty line in old:
+        # line.strip("") checks whether the string is all whitespace
+        while j < len(lines_old) - 1 and not lines_old[j].strip(" "):
```
pv: Why `strip(" ")` and not `strip()`?

numpydoc/numpydoc.py
```diff
@@ -184,6 +183,31 @@ class NumpyCDomain(ManglingDomainBase, CDomain):
...
 
 
+def match_items(lines, content_old):
+    """Create the right items for lines.
+
+    This tries to recover where a line in ``lines``
+    came from before mangling.
+
+    It assumes that missing or new lines
+    are always empty.
```
pv: I'm not sure the idea here is really correct, given that the mangling can do other stuff than only adding empty lines.
(Iirc e.g. plot directive insertion).

pv: iow., the docstrings should say that the result can be wrong

numpydoc/numpydoc.py
amueller: good point.

numpydoc/numpydoc.py
amueller: the docstring of this method? Sure.
And yeah, this function is certainly wrong. However, it is correct in most cases. Doing this properly would mean keeping track of all changes that happen in numpydoc, which is a major undertaking, I think.

amueller: Do you have a better heuristic? As I said in the comment, I'm mostly concerned about knowing which method an error is in.

numpydoc/numpydoc.py
```diff
@@ -184,6 +183,33 @@ class NumpyCDomain(ManglingDomainBase, CDomain):
     }
 
 
+def match_items(lines, content_old):
+    """Create the right items for lines.
```
stefanv: I found myself digging into the docutils source to figure out what these "right 'items'" are.  Can we clear up the docstring of this function a bit, describing what it is doing and why?  I see some of that below, but it's not easy to understand without the context of the PR.

amueller: will do. These are really "Best guess".
So the explanation will be "most of the lines are the same after mangling, but we delete and insert some blank lines. This function tries to match up the blank lines. Any other changes, like including figures or reformatting references will lead to wrong line numbers"

amueller: Actually, this has nothing to do with docutils. The ``mangle_docstring`` function changes the lines but not the items. That's pretty much a design error. It should keep track of both.

amueller: I added a really long-winded explanation and a doctest. Better ? ;)

stefanv:
@amueller The new docstring is excellent, thank you!

amueller:
I don't understand the test failure... "cannot import monkey"

stefanv:
No idea!  Let me restart that build.

NelleV:
@amueller Do you mind rebasing? (I'll try to have Stéfan push the green button today, as this patch will be super useful for our students)

amueller:
@NelleV Done :)
And thanks,  hopefully this will benefit a bunch of numpdoc-based projects.
What are your students doing?

amueller:
@NelleV Done :)
And thanks,  hopefully this will benefit a bunch of numpdoc-based projects.
What are your students doing?

NelleV:
They'll be working on matplotlib. I think the first task is going to be documentation changes.
They'll also be working on scikit-image and possibly some documentation changes as well.

We're supposed to have a website, but I don't know where it is yet.

amueller:
"we"? anyhow, I hope this helps. I've been lobbying for matplotlib documentation changes but didn't find any pots of gold yet :(

NelleV:
We being @stefanv and me, but I just hoped on the project a couple of weeks ago. Stéfan has already run this once during the spring, but I wasn't at Berkeley yet. He knows where the source of the website is, while I am still getting up to speed on pretty much everything.

amueller:
thanks for the merge @stefanv, hope this helps.

stefanv:
Thanks for the patch!

PR #63, opened 2016-09-07
bold attributes so we don't have to deal with trailing underscores

### LINKED ISSUE #62: attribute names with trailing underscores are not references
This is more of a question.
In scikit-learn we name out attributes with trailing underscores like
```
Attributes
==========
bla_ : ndarray
   Nothing useful
```
I remember that at some point, we needed to enclose the attribute in backticks to not get an error about undefined references. Then at some point, we didn't need to do that any more. I just upgraded us to numpydoc master, and now I have to add the backticks in a gain.
Were there any intentional changes in that? It's also possible that we hacked this somewhere ourselves, but I can't find the relevant change in the code :(

amueller:
Also looks better.

Fixes #62.

If the bold formatting is not for you, we can do backticks instead. I feel that attributes should be literals, not interpreted by sphinx.

### DIFF
--- a/numpydoc/docscrape_sphinx.py
+++ b/numpydoc/docscrape_sphinx.py
@@ -132,15 +132,15 @@ class SphinxDocString(NumpyDocString):
                 out += [''] + autosum
 
             if others:
-                maxlen_0 = max(3, max([len(x[0]) for x in others]))
-                hdr = sixu("=")*maxlen_0 + sixu("  ") + sixu("=")*10
+                maxlen_0 = max(3, max([len(x[0]) + 4 for x in others]))
+                hdr = sixu("=") * maxlen_0 + sixu("  ") + sixu("=") * 10
                 fmt = sixu('%%%ds  %%s  ') % (maxlen_0,)
                 out += ['', '', hdr]
                 for param, param_type, desc in others:
                     desc = sixu(" ").join(x.strip() for x in desc).strip()
                     if param_type:
                         desc = "(%s) %s" % (param_type, desc)
-                    out += [fmt % (param.strip(), desc)]
+                    out += [fmt % ("**" + param.strip() + "**", desc)]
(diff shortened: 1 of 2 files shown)


rgommers:
That code is called from ``for param_list in ('Attributes', 'Methods'):`` so your change applies to method as well as attribute tables. And methods get turned into links, so I think this will break that (didn't try).

rgommers:
Bold formatting for attributes sounds fine to me, should look OK visually.

amueller:
I'm not sure what you mean by "methods get turned into links". I guess that depends on the template?
Here there are separate pages for methods:
http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.html

Scikit-learn doesn't do that:
http://scikit-learn.org/stable/modules/generated/sklearn.cluster.AffinityPropagation.html#sklearn.cluster.AffinityPropagation

Do you have a minimal example with methods that are links so I can test? I imagine that if a method contains a trailing underscore, that's also a problem.

pv:
Probably it's about the method autosummary:: table immediately below the 
heading, which contains the links.

rgommers:
No I meant actual ``Methods`` sections in the docstring, like here:
https://github.com/scipy/scipy/blob/master/scipy/stats/kde.py#L80
http://scipy.github.io/devdocs/generated/scipy.stats.gaussian_kde.html#scipy.stats.gaussian_kde

Just tested this PR, looks like things still work so no issue.

rgommers:
OK, in it goes. Thanks @amueller

amueller:
thanks. Sorry for being unresponsive.

rgommers:
> thanks. Sorry for being unresponsive.

Eh? A couple of days doesn't count as unresponsive in my book. Or otherwise I should go make many handfuls of apologies right now:)

PR #65, opened 2016-09-13
Raises an error when a section is present twice

### LINKED ISSUE #64: Having section "Notes" twice does not raise a warning, but only renders once
A lot of matplotlib's plotting function have a notes section concatenated to them at runtime. That leads to docstrings such as the following:

(from plt.imshow)
```
blah blah blah

See also
--------
matshow : Plot a matrix or an array as an image.

Notes
-----
Unless *extent* is used, pixel centers will be located at integer
coordinates. In other words: the origin will coincide with the center
of pixel (0, 0).

Examples
--------

.. plot:: mpl_examples/pylab_examples/image_demo.py

Notes
-----

In addition to the above described arguments, this function can take a
**data** keyword argument. If such a **data** argument is given, the
following arguments are replaced by **data[<arg>]**:

* All positional and all keyword arguments.

Additional kwargs: hold = [True|False] overrides default hold state
"""
```

The rendered docstring is here: http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.imshow
You can notice that the second "Notes" replaced the first one.
And it does this without raising a warning.

NelleV:
The previous behavior overwrote the content of the first section with the second.

Now, it raises a value error.

I also added some elements to gitignore for vim users.

closes #64

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -327,7 +327,7 @@ class NumpyDocString(collections.Mapping):
             if not section.startswith('..'):
                 section = (s.capitalize() for s in section.split(' '))
                 section = ' '.join(section)
-                if self[section]:
+                if self.get(section):
                     msg = ("The section %s appears twice in the docstring." %
                            section)
                     raise ValueError(msg)
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -327,6 +327,11 @@ class NumpyDocString(collections.Mapping):
             if not section.startswith('..'):
                 section = (s.capitalize() for s in section.split(' '))
                 section = ' '.join(section)
+                if self[section]:
+                    msg = ("The section %s appears twice in the docstring." %
+                           section)
+                    raise ValueError(msg)
+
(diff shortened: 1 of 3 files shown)


numpydoc/docscrape.py
```diff
@@ -327,6 +327,11 @@ def _parse(self):
             if not section.startswith('..'):
                 section = (s.capitalize() for s in section.split(' '))
                 section = ' '.join(section)
+                if self[section]:
```
pv: self.get(section)?

NelleV: Why would that be better? If `section` is not a key of the dictionary (nor starts with `..`), an error will be raise later on (l.344). But maybe I am missing something.

pv: you dont want keyerror here for unknown sections

14.9.2016 0.07 "Nelle Varoquaux" <notifications@github.com> kirjoitti:

NelleV:
Thanks @pv for the review! I've fixed the issue.

NelleV:
Thanks!

PR #71, opened 2016-11-04
Remove non Python3 compatible files

### LINKED ISSUE #8: compiler module deprecated
Currently raises `ImportError` on Python 3.

As mentioned [here](http://stackoverflow.com/questions/909092/why-is-the-compiler-package-discontinued-in-python-3), the module's functionality should be replaced with one of these:
- [compile](http://docs.python.org/3/library/functions.html#compile)
- [ast](http://docs.python.org/3.3/library/ast.html)

### LINKED ISSUE #35: PendingDeprecationWarning in Python 3.4
The `imp` module will be deprecated in favor of `importlib`. The module is imported in
- numpydoc/plot_directive.py
- numpydoc/phantom_import.py

### LINKED ISSUE #70: Some files are not compatible with Python 3
In our case the numpydoc functionality does not seem to be affected as these files don't seem to be used. 

Super hacky one-liner to get the full list of problematic files:

```
for i in numpydoc/*.py; do ipython -c "import `echo ${i/.py/} | sed -e 's@/@.@'`" | grep Traceback > /dev/null || echo $i; done
```
Output:
```
numpydoc/comment_eater.py
numpydoc/compiler_unparse.py
numpydoc/phantom_import.py
numpydoc/plot_directive.py
numpydoc/traitsdoc.py
```

Most of the files are 4 years old but traitsdoc.py which is 2 years old. Are they still being used/maintained?

One I am particularly surprised by is plot_directive.py, is it really needed and how is it different from the matplotlib one?

lesteve:
Quoting @pv in #70:
> I believe these file are not widely used, so porting them to Python 3 may not be worthwhile. I don't think deleting them will cause a large inconvenience for anyone.

Fixes #70.
Closes #8. Closes #35.


lesteve:
I quickly looked for numpydoc traitsdoc via the github web UI and could not find any usage outside that was not test_traitsdoc.py same thing for plot_directive. I am happy to search more systematically via the github API if that's deemed necessary..

rkern:
I think all uses of `traitsdoc` have been vendorized, so it is not necessary to distribute it with `numpydoc`.

stefanv:
Thanks, Robert.  Merging.

Issue #80, opened 2016-12-13
using numpydoc resulting in autosummary import error.


SirNO:
When not explicitly declaring a method section in the class docstring numpydoc will create a autosummary table containing all public methods as an entry.
Also when only the methods names are contained in the section, i.e. `method` instead of `method()` numpydoc will also create an autosummary:
```
.. autosummary::

    method1
    method2
    <...>
```

This behaviour would be amazing, but it doesn't work unfortunately because autosummary will be unable to import the methods and thus create a table not containing any descriptions. 
This may be an issue with autosummary, either on sphinx side or because the table is not generated properly. It may also be that the way I documented things is unintended. Anyways, here is a small example producing exactly the issue I am having. I'd be glad about any advice, thanks.

Minimal example:
[example.zip](https://github.com/numpy/numpydoc/files/649095/example.zip)

rgommers:
> It may also be that the way I documented things is unintended.

I don't think it's unintended, but what you're doing does look different from what numpy, scipy & co are doing (which is to have a function listing per submodule, so then you'd be in the right ``currentmodule`` already probably.

Browsing your example, you do seem to need the Child subclass to demonstrate the problem. Your problem description above doesn't mention this only happening for subclasses. I have the feeling your problem summary is a bit too general.

SirNO:
Oh yes, indeed it is a bit general, sorry.

The problem only occurs when using `.. autoclass::` and not when using `.. automodule::`. I didn't notice.

Also, I don't need the childclass to demonstrate the problem, it was just the way I produced the problem in the first place. When you remove the summary string (and the brackets) of `aMethod` in the methods section of `Base`, import will fail:

`/example/docs/src/lib.classes/Base.rst:15: WARNING: failed to import aMethod`

It does not fail before, because `aMethod()` (with brackets) can not be imported either way, so numpydoc generates a table directly from the docstring.

rgommers:
gh-81 merged, so closing. Thanks @SirNO

PR #81, opened 2016-12-13
Added ..currentmodule:: before ..autosummary::

SirNO:
This addition solves an issue I had with the method section generated by numpydoc

### DIFF
--- a/numpydoc/docscrape_sphinx.py
+++ b/numpydoc/docscrape_sphinx.py
@@ -126,6 +126,9 @@ class SphinxDocString(NumpyDocString):
                     others.append((param, param_type, desc))
 
             if autosum:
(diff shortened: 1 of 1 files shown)


SirNO:
this resolves #80

stefanv:
Can anyone else confirm this issue?

rgommers:
I've confirmed that this fixes the example given in gh-80 (which looks like a valid way to use ``autoclass``) and that it doesn't cause any problems for the scipy doc build. So merging.

Thanks @SirNO

rgommers:
Hmm, this passed its tests earlier but is now broken after merging says TravisCI. Will investigate.

rgommers:
Fix in gh-86. @SirNO can you please double check that?

PR #83, opened 2016-12-27
BUG: fix installing numpydoc, broken with missing templates after gh-77.

rgommers:


rgommers:
Merging this straight away, since master is currently not installable.

PR #84, opened 2016-12-27
TST: run test suite after installing via sdist.

rgommers:
This should catch issues like the one fixed in gh-83.
``python setup.py test`` should not be used, it's bad practice.


rgommers:
Checked the build log, this works as intended.

rgommers:
Simple fix and open for long enough, so merging.

PR #85, opened 2016-12-27
MAINT: set minimum Sphinx version to 1.2.3, remove linkcode extension.

rgommers:
linkcode has been included in Sphinx 1.2, and has been giving a
warning when running the numpydoc tests for a long time.

We test with Sphinx 1.2.3 and latest; no need to keep supporting
even older versions than 1.2.3

EDIT: closes gh-85.


rgommers:
Open for long enough, merging.

PR #86, opened 2016-12-27
TST: update test for change in gh-81.

rgommers:


SirNO:
Yes, tests as intended now. 

I will test more carefully in the future. Thanks for the fix.

rgommers:
Thanks for checking

> I will test more carefully in the future. 

No worries, not your fault. It was strange behavior of the CI system, the same test passed when you submitted your PR.

PR #91, opened 2017-04-04
BUG: Make docscrape ParseError Python 3 compatible.

ehebert:
Fix `__str__` for `ParseError` in Python 3, which was resulting in error
messages with `numpydoc.docscrape.ParseError: <unprintable ParseError object>`
instead of the intended descriptive message.

Use `args[0]` instead of `Exception`'s `message` member, which had been removed
in Python 3. `args` is available to both Python 2 and 3. The `__str__` method
had been failing on an `AttributeError` when using `self.message`, resulting in
an unprintable Exception, which occluded the offending typo in a function name
which triggered the`ParseError`.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -91,7 +91,7 @@ class Reader(object):
 
 class ParseError(Exception):
     def __str__(self):
-        message = self.message
+        message = self.args[0]
         if hasattr(self, 'docstring'):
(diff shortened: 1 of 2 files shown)


stefanv:
I'm probably not the right person to review this.  @carreau, can you give a hand?

rgommers:
LGTM, merged. Thanks @ehebert

PR #93, opened 2017-05-10
Fix regression in #81

jschueller:
It seems #81 introduced some kind of bug:

I have the following rst file documenting MyClass object from ottemplate module:
```
User manual
===========

Reference
---------

.. currentmodule:: ottemplate

.. autosummary::
    :toctree: _generated/
    :template: class.rst_t

    MyClass
```

with class.rst_t:
```
{{ objname }}
{{ underline }}

.. currentmodule:: {{ module }}

.. autoclass:: {{ objname }}

   {% block methods %}
   .. automethod:: __init__
   {% endblock %}

```
results in the following error:
```
following exception was raised:
  File "/home/schueller/.local/lib/python3.5/site-packages/sphinx/ext/autodoc.py", line 551, in import_object
    __import__(self.modname)
ImportError: No module named 'ottemplate.ottemplate.MyClass'; 'ottemplate.ottemplate' is not a package

```
I use sphinx  latest stable 1.5.5 with python 3.5.2.

What do you think @SirNO ?

### DIFF
--- a/numpydoc/docscrape_sphinx.py
+++ b/numpydoc/docscrape_sphinx.py
@@ -137,9 +137,6 @@ class SphinxDocString(NumpyDocString):
                     others.append((param, param_type, desc))
 
             if autosum:
-                out += ['.. currentmodule:: '
-                        + self._obj.__module__ + '.' + self._obj.__name__]
-                out += []
                 out += ['.. autosummary::']
                 if self.class_members_toctree:
                     out += ['   :toctree:']
--- a/numpydoc/tests/test_docscrape.py
+++ b/numpydoc/tests/test_docscrape.py
@@ -904,8 +904,6 @@ def test_class_members_doc_sphinx():
 
     .. rubric:: Attributes
 
-    .. currentmodule:: test_docscrape.Foo
-
     .. autosummary::
        :toctree:
 


SirNO:
Hey, 

could it be that the double use of .. currentmodule:: causes the issue? your template 
already sets the module, but then again this should also fail when someone wants to
autodoc multiple classes in the same .rst file, which would be an issue.

idk, back then i just knew i needed to include that line and searched for some way to 
put it there. if you want i can take a closer look tomorrow.

jschueller:
Hi, thanks for your quick reply

I tried to get rid of my currentmodule but it doesnt work either.

It seems it's more complicated than that ; MyClass lies in ```ottemplate/ottemplate.py``` but in ```ottemplate/__init__.py``` I do ```from .ottemplate import *``` so the code location (ottemplate.ottemplate.MyClass) is different from the import location (ottemplate.MyClass). Then your code actually uses the code location whereas it should use the import location.

jschueller:
ping @SirNO

SirNO:
hey, sorry i was quite busy and am without computer the next couple of days.

however, it shouldnt be a problem that you packaged like this since ottemplate.ottemplate.MyClass should still be directly importable from the absolute path.

either way. i think reverting is a good plan. i can easily fix my own stuff by manually adding the line
and it will probably will cause more headaches for many people once there is a release.

so if possible somebody should merge this pr before the next release.

jschueller:
hello maintainers (@rgommers @pv, @stefanv ?), could you review this one ?

jschueller:
ping

stefanv:
This was merged by @rgommers.  Also see gh-86.

jschueller:
@stefanv, no: #86 was meant to fix a single test introduced after #81, this basically reverts #81 as it introduced a severe regression

stefanv:
@jschueller Understood; just tying together all relevant PRs.

rgommers:
Looks like this makes sense to revert, thanks @jschueller, all

PR #95, opened 2017-05-18
add context to signature mangling errors

amueller:
This is slightly hacky (again) but hopefully also useful.
I ran into #67 in scikit-learn, but there was no info whatsoever where this came from. Now it gives the object name and file, so I could actually fix it. Even if #67 gets merged, there is still an error about the having yield and return which would have no context. If that is also removed then this code is not needed any more.

### DIFF
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -113,8 +113,11 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
     if not hasattr(obj, '__doc__'):
         return
-
-    doc = SphinxDocString(pydoc.getdoc(obj))
+    try:
+        doc = SphinxDocString(pydoc.getdoc(obj))
+    except ValueError as e:
+        filename = obj.__code__.co_filename
+        raise ValueError(str(e) + " in {}:{}".format(filename, name))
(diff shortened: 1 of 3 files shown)


amueller:
As an FYI: we accidentally had duplicate sections and they should have just been fixed, but without this patch it's hard to figure out where.

numpydoc/numpydoc.py
```diff
@@ -113,8 +113,11 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
     if not hasattr(obj, '__doc__'):
         return
-
-    doc = SphinxDocString(pydoc.getdoc(obj))
+    try:
+        doc = SphinxDocString(pydoc.getdoc(obj))
+    except ValueError as e:
+        filename = obj.__code__.co_filename
```
pv: That's Python 3 only, needs to be:
```
try:
    filename = inspect.getsourcefile(obj)
except TypeError:
    filename = None
```

pv: Also, the full object name should in principle contained in `name` --- does it get shown by Sphinx?

amueller:
@pv thanks, I now put the fix in the right place and made this much less hacky. That allowed me to add a test.

pv:
thanks lgtm

PR #97, opened 2017-06-18
Set version 0.7.0.dev0

jschueller:


numpydoc/__init__.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import division, absolute_import, print_function
 
-__version__ = '0.6.0'
+__version__ = '0.7.dev'
```
rgommers: Good idea. This is not a valid PEPxxx (forgot the spec number) version number though - can you please make it ``0.7.0.dev0`` ?

jschueller:
ok

rgommers:
Thanks, squashed and merged in 13ae28b

PR #100, opened 2017-06-20
REL: get README content on PyPI front page for new releases

rgommers:


pv:
In principle should use context manager to do `with open(...) as f: 
...`, but probably doesn't matter that much in practice here.

rgommers:
>  but probably doesn't matter that much in practice here.

true, but good point anyway - fixed

pv:
Rebased and merged.

PR #149, opened 2017-11-22
DON'T MERGE: Changes for ODL

kohr-h:
This is just for viewing. Our own addition is replayed on top of the current upstream master. I removed the special (double underscore) methods from the list of public methods.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -555,7 +555,16 @@ class FunctionDoc(NumpyDocString):
 
 class ClassDoc(NumpyDocString):
 
-    extra_public_methods = ['__call__']
+    extra_public_methods = ['_call',
+                            '_apply',
+                            '_lincomb',
+                            '_multiply',
+                            '_divide',
+                            '_dist',
+                            '_norm',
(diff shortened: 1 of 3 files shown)


numpydoc/docscrape.py
```diff
@@ -555,7 +555,16 @@ def __str__(self):
 
 class ClassDoc(NumpyDocString):
 
-    extra_public_methods = ['__call__']
+    extra_public_methods = ['_call',
```
kohr-h: Actually we can remove `_call`, too, since most operators don't provide useful information there anyway.

numpydoc/docscrape.py
```diff
@@ -555,7 +555,16 @@ def __str__(self):
...
+                            '_apply',
+                            '_lincomb',
+                            '_multiply',
+                            '_divide',
+                            '_dist',
+                            '_norm',
+                            '_inner']
+
+    excluded_methods = ['next']
```
kohr-h: obsolete?

kohr-h:
Wrong place for the PR, sorry

PR #150, opened 2017-12-30
Add cross-reference links to parameter types

has2k1:
Tokens of the type description that are determined to be "link-worthy"
are enclosed in a new role called `xref_param_type`. This role when
when processed adds a `pending_xref` node to the DOM. If these types
cross-references are not resolved when the build ends, sphinx does
not complain. This forgives errors made when deciding whether tokens
are "link-worthy". And provided text from the type description is not
lost in the processing, the only unwanted outcome is a type link (due
to coincidence) when none was desired.

Added two options:

   1. `numpydoc_xref_param_type`
   2. `numpydoc_xref_aliases`

closes #57

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -42,6 +42,44 @@ numpydoc_use_blockqutoes : bool
   Until version 0.8, parameter definitions were shown as blockquotes, rather
   than in a definition list.  If your styling requires blockquotes, switch
   this config option to True.  This option will be removed in version 0.10.
+numpydoc_xref_param_type : bool
+  Whether to create cross-references for the parameter types in the
+  ``Parameters``, ``Other Parameters``, ``Returns`` and ``Yields``
+  sections of the docstring.
+  ``True`` by default.
+numpydoc_xref_aliases : dict
+  Mappings to fully qualified paths (or correct ReST references) for the
+  aliases/shortcuts used when specifying the types of parameters.
+  Together with the ``intersphinx`` extension, you can map to links
+  in any documentation.
+  The default is an empty ``dict``.
+
+  If you have the following ``intersphinx`` namespace configuration::
+
+      intersphinx_mapping = {
+          'python': ('https://docs.python.org/3/', None),
+          'numpy': ('https://docs.scipy.org/doc/numpy', None),
+      }
+
+  A useful ``dict`` may look like the following::
+
+      numpydoc_xref_aliases = {
+          # python
+          'sequence': ':term:`python:sequence`',
+          'iterable': ':term:`python:iterable`',
+          'string': 'str',
+          # numpy
+          'array': '~numpy.array',
+          'dtype': '~numpy.dtype',
+          'ndarray': '~numpy.ndarray',
+          'matrix': 'numpy.matrix',
+          'array-like': ':term:`numpy:array_like`',
+          'array_like': ':term:`numpy:array_like`',
+      }
+
+   This option depends on the ``numpydoc_xref_param_type`` option
+   being ``True``.
+
 numpydoc_edit_link : bool
   .. deprecated:: edit your HTML template instead
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -80,6 +80,15 @@ numpydoc_xref_aliases : dict
    This option depends on the ``numpydoc_xref_param_type`` option
    being ``True``.
 
+numpydoc_xref_ignore : set
+    Words not to cross-reference. Most likely, these are common words
+    used in parameter type descriptions that may be confused for
+    classes of the same name. For example::
+
+        numpydoc_xref_ignore = {'type', 'optional', 'default'}
+
+    The default is an empty set.
+
 numpydoc_edit_link : bool
   .. deprecated:: edit your HTML template instead
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -69,9 +69,9 @@ numpydoc_xref_aliases : dict
           'iterable': ':term:`python:iterable`',
           'string': 'str',
           # numpy
-          'array': '~numpy.array',
-          'dtype': '~numpy.dtype',
-          'ndarray': '~numpy.ndarray',
+          'array': 'numpy.ndarray',
+          'dtype': 'numpy.dtype',
+          'ndarray': 'numpy.ndarray',
           'matrix': 'numpy.matrix',
           'array-like': ':term:`numpy:array_like`',
           'array_like': ':term:`numpy:array_like`',
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -50,8 +50,8 @@ numpydoc_xref_param_type : bool
 numpydoc_xref_aliases : dict
   Mappings to fully qualified paths (or correct ReST references) for the
   aliases/shortcuts used when specifying the types of parameters.
-  Together with the ``intersphinx`` extension, you can map to links
-  in any documentation.
+  The keys should not have any spaces. Together with the ``intersphinx``
+  extension, you can map to links in any documentation.
   The default is an empty ``dict``.
 
   If you have the following ``intersphinx`` namespace configuration::
--- a/numpydoc/xref.py
+++ b/numpydoc/xref.py
@@ -157,7 +157,7 @@ def xref_param_type_role(role, rawtext, text, lineno, inliner,
             elif prefix == '~':
                 title = target.split('.')[-1]
 
-    contnode = nodes.inline(title, title)
+    contnode = nodes.Text(title, title)
     node = addnodes.pending_xref('', refdomain='py', refexplicit=False,
                                  reftype='class', reftarget=target)
     node += contnode
--- a/numpydoc/xref.py
+++ b/numpydoc/xref.py
@@ -1,9 +1,22 @@
 import re
 from sphinx import addnodes
 
+# When sphinx (including the napoleon extension) parses the parameters
+# section of a docstring, it converts the information into field lists.
+# Some items in the list are for the parameter type. When the type fields
+# are processed, the text is split and some tokens are turned into
+# pending_xref nodes. These nodes are responsible for creating links.
+#
+# numpydoc does not create field lists, so the type information is
+# not placed into fields that can be processed to make links. Instead,
+# when parsing the type information we identify tokens that are link
+# worthy and wrap them around a special role (xref_param_type_role).
+# When the role is processed, we create pending_xref nodes which are
+# later turned into links.
+
 
 QUALIFIED_NAME_RE = re.compile(
-    # e.g int, numpy.array, ~numpy.array
+    # e.g int, numpy.array, ~numpy.array, .class_in_current_module
     r'^'
     r'[~\.]?'
     r'[a-zA-Z_]\w*'
@@ -11,25 +24,10 @@ QUALIFIED_NAME_RE = re.compile(
     r'$'
 )
 
-CONTAINER_TYPE_RE = re.compile(
-    # e.g.
-    #  - list[int]
-    #  - dict(str, int)
-    #  - dict[str, int]'
-    #  - tuple(float, float)
-    #  - dict[tuple(str, str), int]'
-    r'^'
-    r'(dict|list|tuple)'
-    r'[\[\(]'
-    r'(.+?(?:,\s*)?)+'
-    r'[\]\)]'
-    r'$'
-)
-
 CONTAINER_SPLIT_RE = re.compile(
     # splits dict(str, int) into
     #    ['dict', '[', 'str', ', ', 'int', ']', '']
-    r'(\s*[\[\]\(\),]\s*)'
+    r'(\s*[\[\]\(\)\{\},]\s*)'
 )
 
 DOUBLE_QUOTE_SPLIT_RE = re.compile(
@@ -38,19 +36,27 @@ DOUBLE_QUOTE_SPLIT_RE = re.compile(
     r'(``.+?``)'
 )
 
-IGNORE = {'of', ' of ', 'either', 'or', 'with', 'in', 'default'}
+ROLE_SPLIT_RE = re.compile(
+    # splits to preserve ReST roles
+    r'(:\w+:`.+?(?<!\\)`)'
+)
+
+TEXT_SPLIT_RE = re.compile(
+    # splits on ' or ', ' | ', ', ' and ' '
+    r'(\s+or\s+|\s+\|\s+|,\s+|\s+)'
+)
+
+IGNORE = {'of', 'either', 'or', 'with', 'in', 'default', 'optional'}
 CONTAINER_CHARS = set('[](){}')
 
 
 def make_xref_param_type(param_type, xref_aliases):
     """
     Enclose str in a role that creates a cross-reference
-
     The role ``xref_param_type`` *may be* added to any token
     that looks like type information and no other. The
     function tries to be clever and catch type information
     in different disguises.
-
     Parameters
     ----------
     param_type : str
@@ -58,7 +64,6 @@ def make_xref_param_type(param_type, xref_aliases):
     xref_aliases : dict
         Mapping used to resolve common abbreviations and aliases
         to fully qualified names that can be cross-referenced.
-
     Returns
     -------
     out : str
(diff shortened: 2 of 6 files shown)


numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
+import re
+from sphinx import addnodes
+
+
+QUALIFIED_NAME_RE = re.compile(
+    # e.g int, numpy.array, ~numpy.array
+    r'^'
+    r'[~\.]?'
```
jnothman: what does beginning with . mean?

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+    r'(\s*[\[\]\(\),]\s*)'
+)
+
+DOUBLE_QUOTE_SPLIT_RE = re.compile(
+    # splits 'callable ``f(x0, *args)`` or ``f(x0, y0, *args)``' into
+    #    ['callable ', '``f(x0, *args)``', ' or ', '``f(x0, y0, *args)``', '']
+    r'(``.+?``)'
+)
+
+IGNORE = {'of', ' of ', 'either', 'or', 'with', 'in', 'default'}
```
jnothman: Should these settings be wrapped in a class, e.g. `NumpyDocXrefSettings` so that the user can do the following in config:

```python
numpydoc_xref_param_type = NumpyDocXrefSettings()
numpydoc_xref_param_type.ignore.update(['hello'])
```
or perhaps
```python
numpydoc_xref_param_type.add_to_ignore(['hello'])
```

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+    # The cases are dealt with in an order the prevents
+    # conflict.
+    # Then the strategy is:
+    #   - Identify a pattern we are not interested in
+    #   - split off the pattern
+    #   - re-apply the function to the other parts
+    #   - join the results with the pattern
+
+    # endswith ', optional'
+    if param_type.endswith(', optional'):
```
jnothman: Does this strictly need to be at the end? is there a reason not to just include `optional` in ignore?

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+        return '%s, optional' % make_xref_param_type(
+            param_type[:-10],
+            xref_aliases)
+
+    # Any sort of bracket '[](){}'
```
jnothman: Why does this need special casing aside from other punctuation? The only thing I see it doing is ignoring the first token (dict, list, tuple)...?

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # Is splittable
+    for splitter in [' or ', ', ', ' ']:
+        if splitter in param_type:
+            return _split_and_apply_str(param_type, splitter)
+
+    return param_type
+
+
+def xref_param_type_role(role, rawtext, text, lineno, inliner,
```
jnothman: why are we adding a new role, when we could just allow the user to specify a role? How does this differ from existing `default_role` candidates?

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+                        start,
+                        make_xref_param_type(param_type[1:-1], xref_aliases),
+                        end)
+
+    # May have an unsplittable literal
+    if '``' in param_type:
+        return _split_and_apply_re(param_type, DOUBLE_QUOTE_SPLIT_RE)
+
+    # Is splittable
+    for splitter in [' or ', ', ', ' ']:
```
jnothman: This should be configurable.

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+        if splitter in param_type:
+            return _split_and_apply_str(param_type, splitter)
```
jnothman: Is there a reason we are doing this recursively for each delimiter rather than a complete tokenization step followed by handling each token?

numpydoc/xref.py
```diff
@@ -0,0 +1,169 @@
...
+            # of the form '[int, float]'
+            for start, end in ['[]', '()', '{}']:
+                if param_type.startswith(start) and param_type.endswith(end):
+                    return '%s%s%s' % (
+                        start,
+                        make_xref_param_type(param_type[1:-1], xref_aliases),
+                        end)
+
+    # May have an unsplittable literal
+    if '``' in param_type:
```
jnothman: Do we not similarly need to handle strings with spaces in them? I don't know where such are used, but I could certainly imagine it.

Also existing markup may include things like ```:ref:`blah blah <foo>` ``` which I don't think are currently handled.

numpydoc/xref.py
has2k1: `.class_in_current_module`

numpydoc/xref.py
has2k1: It can be configurable, however that set really doesn't affect the final outcome. It is just to minimise useless markup from being inserted.

numpydoc/xref.py
has2k1: I tried to be precise in targeting the common patterns. I think loosening up would reduce the code.

numpydoc/xref.py
has2k1: Nice catch.

numpydoc/xref.py
has2k1: I'll change it to a regex (to accommodate the comment below), but I do not think it needs to be configurable. I think if type information used in the scientific python ecosystem gets more complicated, then other parts of this function will be found wanting too.

numpydoc/xref.py
has2k1: The role is an implementation detail (internal use). I'll add a detailed comment about it in the file.

has2k1:
> Have you tried to find similar functionality in other doc builders? How do they specify this kind of thing?

Yes, see [comment](https://github.com/numpy/numpydoc/pull/150/files#diff-f280307b5630684d2c1e6fa34bb4ec74)

has2k1:
Though I think the implementation catches nearly every bit of type information in numpy and scipy docstrings, it would help if someone can build any of the documentation to see it in action. I'm not capable of doing so. That would highlight any performance issues.

jnothman:
>  That would highlight any performance issues.

By performance do you mean computational performance?

I much prefer you not amending commits, so we can see the changes from review to review. We can squash upon merge.

has2k1:
> By performance do you mean computational performance?

Yes. I saw a concern about it by someone from pandas.

Sorry about the squash. Let me correct it.

jnothman:
Regarding a test bed, I'll get Scikit-learn docs rendered at https://github.com/scikit-learn/scikit-learn/pull/10421

jnothman:
So here is rendered scikit-learn API reference: https://16406-843222-gh.circle-artifacts.com/0/doc/modules/classes.html
Here's an example before this PR: https://16387-843222-gh.circle-artifacts.com/0/doc/modules/generated/sklearn.cluster.DBSCAN.html#sklearn.cluster.DBSCAN.

Comparing these two runs in time: 11:40 became 12:50 (~8% increase). Not negligible, but not egregious if the links are helpful.

Rendered docs are also larger: 26.3KB became 28.1KB (7% increase) for [DBSCAN](https://16406-843222-gh.circle-artifacts.com/0/doc/modules/generated/sklearn.cluster.DBSCAN.html#sklearn.cluster.DBSCAN) as a single example. Note that using the custom role places an `<em>` around every term that is not linked. This increases page size, and may affect rendering if use_blockquotes=True, or if stylesheets are not default.

I'm not sure that the numpydoc_xref_aliases is quite working as intended either: aliasing 'string' to 'str' means that 'str' will appear; 'sparse matrix' becomes 'sparse matrix' with the example aliases. I think you need some <display text> in your `:xref...:` markup.

jnothman:
Btw, feel free to copy my PR to scikit-learn so you can play with it, and I'll close mine.

jnothman:
[fixed some typos above]

numpydoc/xref.py
```diff
@@ -0,0 +1,144 @@
...
+
+DOUBLE_QUOTE_SPLIT_RE = re.compile(
+    # splits 'callable ``f(x0, *args)`` or ``f(x0, y0, *args)``' into
+    #    ['callable ', '``f(x0, *args)``', ' or ', '``f(x0, y0, *args)``', '']
+    r'(``.+?``)'
+)
+
+ROLE_SPLIT_RE = re.compile(
+    # splits to preserve ReST roles
+    r'(:\w+:`.+?(?<!\\)`)'
```
jnothman: I haven't checked: is ` allowed in anchor text?

numpydoc/xref.py
```diff
@@ -0,0 +1,144 @@
...
+        Mapping used to resolve common abbreviations and aliases
+        to fully qualified names that can be cross-referenced.
+    Returns
+    -------
+    out : str
+        Text with parts that may be wrapped in a
+        ``xref_param_type`` role.
+    """
+    if param_type in xref_aliases:
+        param_type = xref_aliases[param_type]
```
jnothman: This should perhaps have:
```python
if QUALIFIED_NAME_RE.match(xref_aliases[param_type]):
    return ':xref_param_type:`%s <%s>`' % (xref_aliases[param_type], param_type)
```
... or preprocess `xref_aliases`

numpydoc/xref.py
has2k1: Yes. But that regex will not catch pathological use use  of backslashes and ticks.

has2k1:
Looking at some of the test times, there seems to be some variance. numpydoc changed output html markup, so the `<em>` is redundant.

has2k1:
I've created a [demo](https://has2k1.github.io/param-type-demo/index.html) to show how the cross-references generated for all type strings from some common packages.

For each page, the top row shows the original string and the bottom row shows the cross-referenced result.

jnothman:
Neat!

What's the best way for me to configure an ignored word (e.g. object)? is
there any way to try linking to terms as well?

has2k1:
> What's the best way for me to configure an ignored word (e.g. object)?

Maybe expose the `IGNORED` set as an option. As the final html no longer has markup for tokens that do not get linked, the original motivation for set is not as strong.

> is there any way to try linking to terms as well?

The  [aliases dict](https://has2k1.github.io/param-type-demo/index.html#welcome-to-param-types-s-documentation) can do terms.

jnothman:
yes, I mean to link to any terms matched by default. I suppose we could
read in our glossary if needed.

has2k1:
> I suppose we could read in our glossary if needed

That would not be fool proof, terms can have spaces and we don't deal well with spaces.

jnothman:
btw the regex you appear to be using in your demo is over-generating at
least on a line with ': ' in it; it should require ' : ' at a minimum

jnothman:
The need to prefix module-local names with . is a bit annoying. Is there a
way to make this more automagical?

jnothman:
I think trailing punctuation is currently inhibiting rexognition. e.g.
"array."

has2k1:
The dot-names do look weird when not enclosed in a class role, but I don't think there is room for cleverness.

I saw the trialling punctuation thing, it may not be worth the trouble.

jnothman:
I would like to be able to refer to neighbors.NearestNeighbors, but it
seems that it won't be linked unless I use sklearn.neighbors.NearestNeighbors.
How would you solve this?

There are certainly cases where I would like to be able to link multi-word
expressions... What solutions might be feasible?

I've taken a look at all of scikit-learn's type specs in your demo. It
looks pretty good, though most of the links currently produced are not very
informative (do I really want to link for str, list and dict?? for int and
float and double?). There are a few terms I would like unlinked (object,
type, matrix), and there are many that I would need to manually produce
links for: estimator, estimators, classifiers, Bunch, callable, generator,
cross-validation generator and similar, joblib.Memory, csc_matrix, sparse
matrix, booleans, DecisionTreeRegressor, number, scipy.sparse, RandomState,
pair, record array, seq, indexable, ?function, various specific class names
prefixed by module or not. You've also alerted me to some omissions in our
glossary, and non-standard uses of type specs in docstrings.

has2k1:
To cross-reference strings with spaces like `sparse matrix`, it may be useful to change the documentation to `sparse-matrix`.

Initially I thought the aliases dictionary would be small, but it now looks like we want it to be as instrumental as the Rosetta Stone. For a package like sklearn may be you can generate the
`UpperCaseClassName` alias entries from `__all__` in the init files, you have the benefit of good class naming conventions.

On the whole, the parameter types across the whole scipython ecosystem could do with some attempt at standardisation. For example, entries like `list, length = len(coefs_) + len(intercepts_)`  where the code is not double-quoted are not handled with nicely; those underscores make sphinx complain

```
param-type-demo/source/sklearn-param-types.rst:3601: WARNING: Unknown target name: "coefs".
param-type-demo/source/sklearn-param-types.rst:3601: WARNING: Unknown target name: "intercepts".
```

Luckily those are the only complaints in the whole demo, and I think the related docs are not part of the user API.

jnothman:
I don't think standardisation improves quality enough to justify
maintenance costs. Yes, were those docs being rendered, we would likely
have caught that issue and escaped or enclosed the offending terms.

Are you sure generating the aliases is the best way to go?

has2k1:
> Are you sure generating the aliases is the best way to go?

For sklearn, yes. I think it approaches the auto-magic that you wished for, i.e leaving out the dot. The naming conventions seem to be strictly followed and the classes are imported for the user API in a uniform manner.

has2k1:
The set of words to ignore is now configurable; the [variation](https://github.com/scikit-learn/scikit-learn/pull/10426#issuecomment-356033570) in building time conceals any performance impact; and that checks off my boxes.

As the default is set to enable this feature, maybe a notice to other projects!

CC: @charris, @rgommers, @tacaswell, @jorisvandenbossche

doc/install.rst
```diff
@@ -42,6 +42,53 @@ numpydoc_use_blockqutoes : bool
...
   than in a definition list.  If your styling requires blockquotes, switch
   this config option to True.  This option will be removed in version 0.10.
+numpydoc_xref_param_type : bool
+  Whether to create cross-references for the parameter types in the
+  ``Parameters``, ``Other Parameters``, ``Returns`` and ``Yields``
+  sections of the docstring.
+  ``True`` by default.
+numpydoc_xref_aliases : dict
+  Mappings to fully qualified paths (or correct ReST references) for the
+  aliases/shortcuts used when specifying the types of parameters.
```
jnothman: Mention no spaces allowed

tacaswell:
I am :+1: in principle to the functionality, but don't have the bandwidth to do a code review.

jnothman:
I'm not certain it should be enabled by default, so that's something else
for reviewers to comment on

larsoner:
I have wanted something like this for a while. It would be great to resurrect this.

> I would like to be able to refer to neighbors.NearestNeighbors, but it seems that it won't be linked unless I use sklearn.neighbors.NearestNeighbors. How would you solve this?

The way that the `autolink` role in Sphinx handles this is that it uses the `currentmodule`, and moves up from there. So if you were documenting something in `sklearn.linear_model`, you could call it `neighbors.NearestNeighbors`, as it would search first `sklearn.linear_model.neigbors.NearestNeighbors`, which does not exist, and then move on to `sklearn.neighbors.NearestNeighbors`, which does. 

And ideally a `~neighbors.NearestNeighbors` would also work, stripping out everything before the last `.` so rendering as `NearestNeighbors` in the doc.

Does that sound tractable? Any interest in coming back to this? I have recently been working on the SciPy docs, and maintain docs for MNE-Python, which follows a lot of the same conventions as `sklearn`, so I'd be happy to test this out and review.

larsoner:
> I'm not certain it should be enabled by default

I agree this would be safer as opt-in.

has2k1:
@larsoner, it is already opt-in and the `~` works as expected. I have had this feature working without any issues in [plotnine](https://plotnine.readthedocs.io/en/stable/index.html) for about a year.

larsoner:
Okay, let me try it with MNE and SciPy and see what happens

larsoner:
@has2k1 can you rebase on latest `master` to get rid of the conflicts?

doc/install.rst
```diff
@@ -42,6 +42,53 @@ numpydoc_use_blockqutoes : bool
   Until version 0.8, parameter definitions were shown as blockquotes, rather
   than in a definition list.  If your styling requires blockquotes, switch
   this config option to True.  This option will be removed in version 0.10.
+numpydoc_xref_param_type : bool
+  Whether to create cross-references for the parameter types in the
+  ``Parameters``, ``Other Parameters``, ``Returns`` and ``Yields``
+  sections of the docstring.
+  ``True`` by default.
```
larsoner: If it's opt-in, then this docstring should be changed to say "``False`` by default"

PR #171, opened 2018-04-12
DOC: Add missing backslash to omega for inline math example.

pvanmulbregt:
In doc/example.py capitalize "greek" and add a missing backslash to omega so that it shows up as a symbol rather then the word.


mattip:
LGTM

rgommers:
Merged, thanks @pvanmulbregt

PR #175, opened 2018-05-02
FIX parsing of type-only return params

### LINKED ISSUE #72: Anonymous return values have their types populated in the name slot of the tuple.
I noticed an inconsistency, when using numpydoc version 0.6.0 in python2.7 on Ubuntu. The parsed return section information returns different styles of tuple depending on if the return value is anoymous or not. 

Here is a minimal working example: 

```python

def mwe():
        docstr = (
            'Returns\n'
            '----------\n'
            'int\n'
            '    can return an anoymous integer\n'
            'out : ndarray\n'
            '    can return a named value\n'
        )
        doc = NumpyDocString(docstr)
        returns = doc._parsed_data['Returns']
        print(returns)
```

This results in 

```python
[(u'int', '', [u'can return an anoymous integer']),
 (u'out', u'ndarray', [u'can return a named value'])]
```

However judging by tests (due to lack of docs), I believe it was indented that each value in the returns list should be a tuple of `(arg, arg_type, arg_desc)`. Therefore we should see this instead: 

```python
[('', u'int', [u'can return an anoymous integer']),
 (u'out', u'ndarray', [u'can return a named value'])]
```

My current workaround is this:

```python
        for p_name, p_type, p_descr in returns:
            if not p_type:
                p_name = ''
                p_type = p_name
```

timhoffm:
Fixes #72.

The sections "Parameters" and "Returns" have different semantics for single-element parameter headers. For "Parameters", the header is the name, for "Returns" the header is the type. This was not reflected in the code as both sections used the same parser.

Note: This is just a quick write-up how the code should work. I've not yet let run the tests (they may need fixing as well as mentioned in #72). But for brevity I'll let CI decide on that.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -220,7 +220,7 @@ class NumpyDocString(Mapping):
             else:
                 yield name, self._strip(data[2:])
 
-    def _parse_param_list(self, content):
+    def _parse_param_list(self, content, single_element_is_type=False):
         r = Reader(content)
         params = []
         while not r.eof():
@@ -228,7 +228,10 @@ class NumpyDocString(Mapping):
             if ' : ' in header:
                 arg_name, arg_type = header.split(' : ')[:2]
             else:
-                arg_name, arg_type = header, ''
+                if single_element_is_type:
+                    arg_name, arg_type = '', header
+                else:
+                    arg_name, arg_type = header, ''
 
             desc = r.read_to_next_unindented_line()
             desc = dedent_lines(desc)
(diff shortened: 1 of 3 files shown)


timhoffm:
Note: CI currently fails because of pip changes. Should be fixed by #174.

jnothman:
if you've not had to modify any tests, how do we know this affects output?

timhoffm:
As said above, I didn't run any tests so far myself. It's apparent that one parser for "Parameters" and "Returns" cannot get both right. I'm confident that the proposed code change itself is correct. What has to be still shown is that the calling code and tests work with that (it might be that they partly compensate for the original bug). I thought I'd use the tests in CI for that, but CI is already prematurely failing for different reasons.

timhoffm:
Rebased onto master.

*Note:* The tests do currently fail. Waiting for #176 before doing any further changes to fix tests.

timhoffm:
PR updated.

Single element returns params such as:
~~~
Returns
-------
int
    The return value.
~~~
were detected as names. I.e. `int` was considered a name. This logical error has been fixed such that `int` is now a type and the name is empty.

As a consequence, `int` is not formatted bold anymore. This is consistent with the formatting of types in patterns like `x : int` and a prerequisite for type references like ``:class:`MyClass` `` to work in this position.

larsoner:
@timhoffm can you rebase? Then I can take a look and hopefully merge

rgommers:
I've taken the liberty of fixing the merge conflict. The only nontrivial change was deciding where the new heading `Receives` goes; I added it to `'Returns', 'Yields', 'Raises', 'Warns'`.

rgommers:
Okay in it goes. Thanks @timhoff, and sorry it took a while! And thanks @larsoner for reviewing

timhoff:
This is funny.  I am a programmer and a member of github, but you missed adding an “m" to the end of the address.  It should be @timhoffm.  LOL

rgommers:
hmm odd, I thought I tab-completed that. Sorry for the noise @timhoff. And thanks @timhoffm !

Issue #188, opened 2018-08-12
 NumpyDocString doesn't roundtrip well


hmaarrfk:
```python
from numpydoc.tests.test_docscrape import doc_txt
from numpydoc import docscrape
docscrape.NumpyDocString(
    str(docscrape.NumpyDocString(doc_txt)))
```
leads to

```
ParseError: `some`_ is not a item name in 'numpy.multivariate_normal(m 
[...]
```

This would be really useful if we want to implement merging (or adding) NumpyDocStrings

PR #189, opened 2018-08-12
Better round tripping for `See also` section

hmaarrfk:
Potential Fix for #188

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -442,7 +442,7 @@ class NumpyDocString(collections.Mapping):
             elif func_role:
                 link = ':%s:`%s`' % (func_role, func)
             else:
-                link = "`%s`_" % func
+                link = "%s" % func
             if desc or last_had_desc:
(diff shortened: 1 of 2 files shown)


pv:
-1, they are meant to become links, and this breaks SphinxDocString behavior.

I'm not exactly sure if what you are trying to achieve aligns with numpydoc goals.
In particular, round-tripping the docstrings is not a design goal for numpydoc.

hmaarrfk:
ok

hmaarrfk:
@pv, could it be provided as an additional function? There has been some interest in having wrapper modify the docstrings:

https://github.com/numpy/numpydoc/pull/67#issuecomment-254192042

Something I wrote:
https://github.com/hmaarrfk/deprecation_factory/blob/cfe72994e991e9833b7084419d5341139a6da97b/deprecation_factory/default_parameter_change.py#L146

Having the ability to roundtrip docstrings would be useful in these cases.

PR #221, opened 2019-05-29
BUG: Defer to autodoc for signatures 

thequackdaddy:
xref #220

I think this should pass. It fails because method signatures include `self`. 

I'll be happy to help make this pass.

### DIFF
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -206,7 +206,8 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
     sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
     if sig:
         sig = re.sub("^[^(]*", "", sig)
-        sig = re.sub("$", "", sig)
+        sig = re.sub(r"\$self,\s", "", sig)
+        sig = re.sub(r"\/,\s", "", sig)
         return sig, ''
 
 
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -202,11 +202,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
     if not hasattr(obj, '__doc__'):
         return
-
-    cfg = {'use_autodoc_signature': app.config.numpydoc_use_autodoc_signature,
-           'show_class_members': False}
-
-    doc = get_doc_object(obj, config=cfg)
+    doc = get_doc_object(obj, config={'show_class_members': False})
     sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
     if sig:
         sig = re.sub("^[^(]*", "", sig)
@@ -230,7 +226,6 @@ def setup(app, get_doc_object_=get_doc_object):
     app.add_config_value('numpydoc_edit_link', None, False)
     app.add_config_value('numpydoc_use_plots', None, False)
     app.add_config_value('numpydoc_use_blockquotes', None, False)
-    app.add_config_value('numpydoc_use_autodoc_signature', False, False)
     app.add_config_value('numpydoc_show_class_members', True, True)
     app.add_config_value('numpydoc_show_inherited_class_members', True, True)
     app.add_config_value('numpydoc_class_members_toctree', True, True)
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -203,8 +203,8 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
     if not hasattr(obj, '__doc__'):
         return
     doc = get_doc_object(obj, config={'show_class_members': False})
-    sig = (doc['Signature'] or
-           _clean_text_signature(getattr(obj, '__text_signature__', None)))
+    sig = (doc['Signature']
+           or _clean_text_signature(getattr(obj, '__text_signature__', None)))
     if sig:
         sig = re.sub("^[^(]*", "", sig)
         return sig, ''
@@ -213,12 +213,11 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 def _clean_text_signature(sig):
     if sig is None:
         return None
-    start_pattern = re.compile(r"^[^(]*\(*")
+    start_pattern = re.compile(r"^[^(]*\(")
     start, end = start_pattern.search(sig).span()
     start_sig = sig[start:end]
-    sig = sig[end:]
-    sig = re.sub(r"\)", "", sig)
-    params = sig.replace(' ', '').split(',')
+    sig = sig[end:-1]
+    params = sig.split(', ')
     for param in ('$self', '$module', '$type', '/'):
         if param in params:
             params.remove(param)
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -203,12 +203,28 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
     if not hasattr(obj, '__doc__'):
         return
     doc = get_doc_object(obj, config={'show_class_members': False})
-    sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
+    sig = (doc['Signature'] or
+           _clean_text_signature(getattr(obj, '__text_signature__', None)))
     if sig:
         sig = re.sub("^[^(]*", "", sig)
-        sig = re.sub(r"\$self,\s", "", sig)
-        sig = re.sub(r"\/,\s", "", sig)
         return sig, ''
+
+
+def _clean_text_signature(sig):
+    if sig is None:
+        return None
+    sig = re.sub(r"^[^(]*\(", "", sig)
+    sig = re.sub(r"\)", "", sig)
+    params = sig.replace(' ', '').split(',')
+    if '$self' in params:
+        params.remove('$self')
+    if '$module' in params:
+        params.remove('$module')
+    if '$type' in params:
+        params.remove('$type')
+    if '/' in params:
+        params.remove('/')
+    return '(' + ', '.join(params) + ')'
 
 
 def setup(app, get_doc_object_=get_doc_object):
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -202,7 +202,11 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 
     if not hasattr(obj, '__doc__'):
         return
-    doc = get_doc_object(obj, config={'show_class_members': False})
+
+    cfg = {'use_autodoc_signature': app.config.numpydoc_use_autodoc_signature,
+           'show_class_members': False}
+
+    doc = get_doc_object(obj, config=cfg)
     sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
     if sig:
         sig = re.sub("^[^(]*", "", sig)
@@ -226,6 +230,7 @@ def setup(app, get_doc_object_=get_doc_object):
     app.add_config_value('numpydoc_edit_link', None, False)
     app.add_config_value('numpydoc_use_plots', None, False)
     app.add_config_value('numpydoc_use_blockquotes', None, False)
+    app.add_config_value('numpydoc_use_autodoc_signature', False, False)
     app.add_config_value('numpydoc_show_class_members', True, True)
     app.add_config_value('numpydoc_show_inherited_class_members', True, True)
     app.add_config_value('numpydoc_class_members_toctree', True, True)
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -217,11 +217,9 @@ def _clean_text_signature(sig):
     start, end = start_pattern.search(sig).span()
     start_sig = sig[start:end]
     sig = sig[end:-1]
-    params = sig.split(', ')
-    for param in ('$self', '$module', '$type', '/'):
-        if param in params:
-            params.remove(param)
-    return start_sig + ', '.join(params) + ')'
+    sig = re.sub(r'^\$(self|module|type)(,\s|$)','' , sig, count=1)
+    sig = re.sub(r'(^|(?<=,\s))/,\s\*', '*', sig, count=1)
+    return start_sig + sig + ')'
 
 
 def setup(app, get_doc_object_=get_doc_object):
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -213,18 +213,16 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
 def _clean_text_signature(sig):
     if sig is None:
         return None
-    sig = re.sub(r"^[^(]*\(", "", sig)
+    start_pattern = re.compile(r"^[^(]*\(*")
+    start, end = start_pattern.search(sig).span()
+    start_sig = sig[start:end]
+    sig = sig[end:]
     sig = re.sub(r"\)", "", sig)
     params = sig.replace(' ', '').split(',')
-    if '$self' in params:
-        params.remove('$self')
-    if '$module' in params:
-        params.remove('$module')
-    if '$type' in params:
-        params.remove('$type')
-    if '/' in params:
-        params.remove('/')
-    return '(' + ', '.join(params) + ')'
+    for param in ('$self', '$module', '$type', '/'):
+        if param in params:
+            params.remove(param)
+    return start_sig + ', '.join(params) + ')'
 
 
 def setup(app, get_doc_object_=get_doc_object):
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -206,6 +206,7 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
     sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
     if sig:
         sig = re.sub("^[^(]*", "", sig)
+        sig = re.sub("$", "", sig)
         return sig, ''
 
 
--- a/numpydoc/tests/test_docscrape.py
+++ b/numpydoc/tests/test_docscrape.py
@@ -1484,6 +1484,35 @@ def test_xref():
     line_by_line_compare(str(doc), xref_doc_txt_expected)
 
 
+def test_signature_inspect():
+
+    def func1(a):
+        pass
+
+    def func2(a, b):
+        pass
+
+    class klass:
+        def __init__(self, a, b):
+            pass
+
+        def meth1(self, a, b):
+            pass
+
+        def meth2(a, b):
+            pass
+
+        @classmethod
+        def meth3(a, b):
+            pass
+
+    assert get_doc_object(func1)['Signature'] == 'func1(a)'
+    assert get_doc_object(func2)['Signature'] == 'func2(a, b)'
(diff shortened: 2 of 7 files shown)


jnothman:
Why do you think this needs to be an option?

thequackdaddy:
I don't. I just don't know if you want to defer.

Could just delete all that code in that case.

Throwing out ideas.

thequackdaddy:
I should have mentioned, it appears that the numpydoc way of getting signatures on class method is inconsistent between python 2 and python 3. Maybe another reason to just say, "screw it... use autodoc signatures". 

But I don't want to upset the apple cart.

thequackdaddy:
(and who builds docs on python 2 anyway)

bashtage:
I  build latest statsmodels doc with this branch and it look good. Handles both function sigs in docstrings (e.g., from Cython) and removes self when needed. For example, (`OLS`)

image

larsoner:
> Maybe another reason to just say, "screw it... use autodoc signatures".
>
> But I don't want to upset the apple cart.

Do tests pass if you always use autodoc? It would be nice to only have one code path if possible, especially if we know that the non-autodoc one is buggy.

thequackdaddy:
@larsoner I honestly haven't tried that. I'll try to get to it in the next few days...

thequackdaddy:
@larsoner So it was much easier than I thought...

Answer is that if I remove this code, we should revert to auto doc.

https://github.com/numpy/numpydoc/blob/b45eb2b4725ffd536d77fb6995983df4775eefb9/numpydoc/docscrape.py#L575-L590

The only induced failure is this one...

https://github.com/numpy/numpydoc/blob/b45eb2b4725ffd536d77fb6995983df4775eefb9/numpydoc/tests/test_docscrape.py#L723-L731

numpydoc/tests/test_docscrape.py
```diff
@@ -720,17 +720,6 @@ def test_parameters_without_extended_description():
...
 
-def test_escape_stars():
-    signature = str(doc3).split('\n')[0]
-    assert signature == r'my_signature(\*params, \*\*kwds)'
-
-    def my_func(a, b, **kwargs):
-        pass
-
-    fdoc = FunctionDoc(func=my_func)
-    assert fdoc['Signature'] == r'my_func(a, b, \*\*kwargs)'
```
larsoner: This might actually be important. Do things with `*args, **kwargs` render properly now? That's what this is meant to check/ensure. Can we replace this test rather than just remove it? Maybe check that the RST generated has the `*`s escaped properly?

thequackdaddy:
Would you be ok if I made a PR to turn on coveralls codecov? Might be easier to see what's really being tested and where.

larsoner:
Sure, they're a bit buggy but we can configure then to be useful

bashtage:
Published statsmodels devel docs using @thequackdaddy 's branch.  **kwargs looks fine.


https://www.statsmodels.org/devel/generated/statsmodels.tsa.statespace.dynamic_factor.DynamicFactor.html?highlight=dynamicfactor#statsmodels.tsa.statespace.dynamic_factor.DynamicFactor

larsoner:
@thequackdaddy feel free to rebase now that #230 is in and then hopefully you can change the lines marked `XXX` here:

https://github.com/numpy/numpydoc/pull/230/files#diff-6624c1bee5f6adf586fc42bb69eb40b6R48-R54

thequackdaddy:
Rebase is done and I think that might be all we need. Thoughts?

jnothman:
Please update the PR title. It certainly looks like a simple solution.

I'm building scikit-learn docs with this for review at https://github.com/scikit-learn/scikit-learn/pull/14681

thequackdaddy:
@jnothman I changed the title.

jnothman:
I like how this looks in the member methods listing (autosummary). It also affects the signatures shown in method documentation. Here I see things like `get_params(self, deep=True)` become `get_params(deep=True)` which is fine, and `__init__(self, /, *args, **kwargs)` become `__init__($self, /, *args, **kwargs)` in some cases (I'm seeing it in sklearn.base, like [here](https://70103-843222-gh.circle-artifacts.com/0/doc/modules/generated/sklearn.base.TransformerMixin.html#sklearn.base.TransformerMixin) but not elsewhere like [here](https://70103-843222-gh.circle-artifacts.com/0/doc/modules/generated/sklearn.covariance.EmpiricalCovariance.html#sklearn.covariance.EmpiricalCovariance)). Any idea what this `$` is about?  More `$` [here](https://70103-843222-gh.circle-artifacts.com/0/doc/modules/generated/sklearn.gaussian_process.kernels.Hyperparameter.html#sklearn.gaussian_process.kernels.Hyperparameter)

gforsyth:
@seberg -- this PR does fix the issues we were discussing at the dask developers meeting.  If this can be merged and a new release cut then we can unpin `numpydoc` in a bunch of downstream projects that are affected by the double-escaping we were seeing in `autosummary`.

jnothman:
@gforsyth, am I alone in getting those $s rendering, per my comment above (https://github.com/numpy/numpydoc/pull/221#issuecomment-522368100)? I don't know why they are appearing...

gforsyth:
Hey @jnothman -- I `grep`d over some rendered documentation to look for extra `$`s and I haven't come across any when building `dask` or `distributed` docs.  I can try to build `sklearn` later today and see if I hit the same issue.

larsoner:
@thequackdaddy I merged a PR that caused a conflict (sorry about that) -- can you rebase or merge with `master`? After that I can try building docs for a few libraries (including sklearn) to see how things look.

@jnothman okay to merge if all *args/**kwargs I see in functions and methods look okay, and I don't see any stray `$`'s?

thequackdaddy:
@larsoner Rebased.

larsoner:
I rebuild docs for SciPy, scikit-learn, and MNE-Python. SciPy slook good, but `MNE` and `scikit-learn` still have the `$self` problem above that @jnothman mentioned. It seems to occur any time there is a method that is defined by the super class. So even in other modules (statsmodels?) you should be able to observe the same behavior. Any ideas about how to fix it @thequackdaddy ?

larsoner:
Okay @thequackdaddy I added some small changes in #260. If you rebase this now, you should see a `$self` show up, and a test fail accordingly.

Let me know if you want to / can look into it. Otherwise I'll try to find some time to do it.

thequackdaddy:
Well I think I understand the issue. 

I don't fully understand it but sometimes, `MyClass.__init__.__text_signature__` is populated with this odd string... `($self, /, *args, **kwargs)`. 

Does someone understand what that does? It looks like something to do with python's underlying C, but I don't really get it. 

I found this stack overflow which tries to explain it...

https://stackoverflow.com/questions/25847035/what-are-signature-and-text-signature-used-for-in-python-3-4

larsoner:
I wonder if this will make it so that classes defined directly in C that previously had valid signatures no longer will. @jnothman know of a good test case for this?

bashtage:
These results

https://github.com/search?l=Python&q=%22__text_signature__%22&type=Code

suggest `__text_signature__` is a legitimate attribute.

thequackdaddy:
I will note that looking at sphinx, this is used only when pulling that field from `list` class. 

https://github.com/sphinx-doc/sphinx/search?q=__text_signature__&unscoped_q=__text_signature__

I haven't gotten my head around it, but I notice that almost all of the results of @bashtage 's search are unit tests. That suggests to me it is of lesser importance. I guess we could just check if you get a "weird" result or write some code to prevent it from printing when you have a `$` in the signature.

thequackdaddy:
git blame ...

https://github.com/numpy/numpydoc/commit/72803560ea974ec071bf47e075b78ee857e777d0

bashtage:
Some more background on __text_signature__: https://stackoverflow.com/a/50763175/2551705  Still not sure why sometimes the $self is there rather than self.

bashtage:
I think this explains it best:

```
I also took the opportunity to modify the signature slightly. Signatures 
were a little inconsistent about whether they specified the "self" 
parameter or not, so there were some complicated heuristics in 
inspect.Signature about when to keep or omit the first argument.  In the 
new format I made this more explicit: if the first argument starts with 
a dollar sign ("$"), that means "this is a special first argument" (self 
for methods, module for module-level callables, type for class methods 
and __new__).  That removed all the guesswork from inspect.Signature; 
now it works great.  (In case you're wondering: I still use ast.parse to 
parse the signature, I just strip out the "$" first.)
```

https://mail.python.org/pipermail/python-dev/2014-February/132245.html

thequackdaddy:
Sounds like I just need to strip out the `$` in `__text_signature__` if it exists then. I'll do that.

thequackdaddy:
> Is there any chance of $module (probably not, mostly built-ins or lambdas) or $type here?

Easy to check for that as well and replace.

numpydoc/numpydoc.py
```diff
@@ -206,6 +206,8 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
     sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
     if sig:
         sig = re.sub("^[^(]*", "", sig)
+        sig = re.sub(r"\$self,\s", "", sig)
```
bashtage, line 209: I think if the signature only has `$self` and no other arguments this may not be correct since it might look like `($self)` not `($self, )`.

thequackdaddy: Hmm... I think you're right. 

Thinking about it more, maybe I should write a helper function that only us used on the `__text_signature__` field. I think essentially it would be something like... 

```
def clean_text_signature(__text_signature__):
    params = ___text_signature__.split(',')
    params.remove('$self')
    params.remove('/')
    if len(params) > 1:
        params = ' ,'.join(params)
    return '(' + params + ')'
```

bashtage: Sounds good, but probably remove $type, $module too?

thequackdaddy: done

numpydoc/numpydoc.py
```diff
@@ -203,12 +203,30 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
...
+    sig = re.sub(r"\)", "", sig)
+    params = sig.replace(' ', '').split(',')
+    if '$self' in params:
+        params.remove('$self')
+    if '$module' in params:
+        params.remove('$module')
+    if '$type' in params:
+        params.remove('$type')
+    if '/' in params:
+        params.remove('/')
```
larsoner, line 226: Seems cleaner and easier to add new ones with:
```
for param in ('$self', '$module', '$type', '/'):
    if param in params:
        params.remove(param)
```

thequackdaddy: done

numpydoc/numpydoc.py
```diff
@@ -203,12 +203,28 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
...
+           _clean_text_signature(getattr(obj, '__text_signature__', None)))
     if sig:
         sig = re.sub("^[^(]*", "", sig)
         return sig, ''
 
 
+def _clean_text_signature(sig):
+    if sig is None:
+        return None
+    start_pattern = re.compile(r"^[^(]*\(*")
```
jnothman, line 216: I don't think it'll make much difference, but why do you have that second `*`?

numpydoc/numpydoc.py
```diff
@@ -203,12 +203,28 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
...
+    start, end = start_pattern.search(sig).span()
+    start_sig = sig[start:end]
+    sig = sig[end:]
+    sig = re.sub(r"\)", "", sig)
```
jnothman, line 220: what is this for? It doesn't appear to pertain to any of the test cases below.

numpydoc/numpydoc.py
```diff
@@ -203,12 +203,28 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
...
+    params = sig.replace(' ', '').split(',')
```
jnothman, line 221: this is too aggressive. Why do we need to replace all spaces? What if the value is rendered with spaces in it?

Doesn't
```suggestion
    params = sig.split(', ')
```
suffice?

numpydoc/tests/test_numpydoc.py
```diff
@@ -64,6 +64,20 @@ def test_mangle_docstrings():
...
+    assert (_clean_text_signature('func($self, *args, **kwargs)')
+            == 'func(*args, **kwargs)')
+    assert _clean_text_signature('($self)') == '()'
+    assert _clean_text_signature('()') == '()'
+    assert _clean_text_signature('func()') == 'func()'
+    assert (_clean_text_signature('func($self, /, *args, **kwargs)')
+            == 'func(*args, **kwargs)')
+    assert _clean_text_signature('($module)') == '()'
+    assert _clean_text_signature('func($type)') == 'func()'
+
```
jnothman, line 79: please add a test where there are other params with default values, such as
```
_clean_text_signature('func($self, foo="hello world")')
_clean_text_signature('func(foo="hello world")')
_clean_text_signature('func(foo="$self")')
```
and tricky cases like
```
_clean_text_signature('func(self, other)')
```

thequackdaddy: done

numpydoc/numpydoc.py
thequackdaddy: Fair point. I wasn't considering default values. Additionally, I think it should be pretty safe to assume that you'd rarely see something like `func(param1,param2)` (with no space) or `func(param1,  param2)` with too many spaces.

numpydoc/numpydoc.py
thequackdaddy: I'm trying to first make it so 'func(param1, param2, ..., paramN)` -> `param1, param2, ..., paramN`. I suppose I could get rid of the last `re.sub` and just do `sig = sig[end:-1]`, but I wasn't confident if there would be trailing spaces.

numpydoc/numpydoc.py
thequackdaddy: fair point. removed.

numpydoc/tests/test_numpydoc.py
```diff
@@ -64,6 +64,32 @@ def test_mangle_docstrings():
...
+    assert (_clean_text_signature("func($self, foo='hello world')")
+            == "func(foo='hello world')")
+    assert (_clean_text_signature('func(foo="hello world")')
+            == 'func(foo="hello world")')
+    assert (_clean_text_signature('func(foo="$self")')
+            == 'func(foo="$self")')
+    assert (_clean_text_signature('func($self, foo="$self")')
+            == 'func(foo="$self")')
+    assert _clean_text_signature('func(self, other)') == 'func(self, other)'
+    assert _clean_text_signature('func(self, $self)') == 'func(self)'
```
jnothman, line 90: I'm not sure this is a thing :)

numpydoc/numpydoc.py
```diff
@@ -203,12 +203,27 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
...
+def _clean_text_signature(sig):
+    if sig is None:
+        return None
+    start_pattern = re.compile(r"^[^(]*\(")
+    start, end = start_pattern.search(sig).span()
+    start_sig = sig[start:end]
+    sig = sig[end:-1]
+    params = sig.split(', ')
+    for param in ('$self', '$module', '$type', '/'):
+        if param in params:
```
jnothman, line 222: I don't think it'll happen often, but I'm still not comfortable with the fact that this processes all parts of the sig. Can't we check explicitly that the `/` precedes a `*`, and that the `$` is at the beginning, rather than use the split and join logic?
```
sig = re.sub(r'^\$(self|module|type), ', '', sig, count=1)
sig = re.sub(r', /, \*', ', *', sig, count=1)
```
Is this not sufficient?

thequackdaddy: I'll confess that I have (almost) no understanding of how these C text signature things work. 

What you have would make it so `func($self)` would still be `func($self)`. Is that alright? I have no idea what the possible syntax is for this.

If you (or someone else) has some kind of documentation on what is possible, that would make this back-and-forth much easier.

jnothman: Sorry, I meant to apply those substitutions after
```
    start_pattern = re.compile(r"^[^(]*\(")
    start, end = start_pattern.search(sig).span()
    start_sig = sig[start:end]
    sig = sig[end:-1]
```
They're just a bit more careful than splitting on every comma.

thequackdaddy: Actually I think this makes sense. Let me change that.

To handle `func($self)`, I guess you could just add a `’, ‘` right before your code. Then use regex to strip off a trailing `’, ‘` if it exists. 

Do you think that would work?

jnothman: Sorry, didn't handle that case, did I?
Can use `r'^\$(self|module|type)(, |$))?'` which will strip a comma, or check that we're at the end.

thequackdaddy: Take a peek at this. I had to do something to check whether the `/` was the start of the string or preceded by a `, `.

numpydoc/numpydoc.py
```diff
@@ -203,12 +203,29 @@ def mangle_signature(app, what, name, obj, options, sig, retann):
...
+    start_pattern = re.compile(r"^[^(]*\(")
+    start, end = start_pattern.search(sig).span()
+    start_sig = sig[start:end]
+    sig = sig[end:-1]
+    sig = re.sub(r'^\$(self|module|type)(,\s|$)','' , sig, count=1)
+    def slashfunc(matchobj):
+        if matchobj.group(0)[0] == '/':
+            return '*'
+        return ', *'
+    sig = re.sub(r'(^|,\s)/,\s\*', slashfunc, sig, count=1)
```
jnothman, line 225: Here are two options that would suffice without a callback:
```suggestion
    sig = re.sub(r'(^|(?<=,\s))/,\s\*', '*', sig, count=1)
```
or
```suggestion
    sig = re.sub(r'(^|,\s)/(?=,\s\*)', '', sig, count=1)
```
These use lookahead or lookbehind to require but not match some parts.

thequackdaddy: Ahh. Genius.

jnothman:
So one last thing: I think this now stops displaying `self`, `cls`, etc. where it used to in method docs.

I suspect we should at least document this in a change log.

thequackdaddy:
> So one last thing: I think this now stops displaying `self`, `cls`, etc. where it used to in method docs.
> 
> I suspect we should at least document this in a change log.

It looks like this project generally builds changelogs as part of the release cycle, unless I'm missing something? I just see a release notes file that looks like it was built for the 1.0 release. I don't see where I can edit in that?

thequackdaddy:
> > So one last thing: I think this now stops displaying `self`, `cls`, etc. where it used to in method docs.
> > I suspect we should at least document this in a change log.
> 
> It looks like this project generally builds changelogs as part of the release cycle, unless I'm missing something? I just see a release notes file that looks like it was built for the 1.0 release. I don't see where I can edit in that?

Never mind. I guess the file is meant to be cumulative. I'll give it the old college try.

thequackdaddy:
Is everyone content?

jnothman:
Thank you @thequackdaddy!

PR #224, opened 2019-05-31
Py3fy some doctests.

anntzer:


larsoner:
Looks like a straightforward fix, thanks @anntzer

PR #225, opened 2019-05-31
MAINT: doctest and pytest

larsoner:
#224 made me realize we should be doctesting. This adds it.


larsoner:
This one is pretty simple and minor, does not change any user-facing behaviors, and should help check other PRs, so I'll go ahead and merge.

PR #226, opened 2019-05-31
BUG: Fix full rebuilds

larsoner:
I realized that my Sphinx was always doing a full rebuild. It's because with `xref` we modify the config to add some values after the builder is initted. But then this env gets pickled and saved to disk, so on the next sphinx build run it compares the user's config (which does not have these values) to the pickled config (which has the added values) which triggers a full rebuild every time.

This fixes it by instead just checking our set after the user's set when doing the `xref` step. It's a tiny bit less efficient but should not be too much of a slowdown.

### DIFF
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -18,6 +18,7 @@ It will:
 """
 from __future__ import division, absolute_import, print_function
 
+from copy import deepcopy
 import sys
 import re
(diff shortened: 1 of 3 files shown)


NicolasHug:
Can confirm this solves the re-build of scikit-learn!

larsoner:
I'll merge tomorrow unless there are objections

PR #234, opened 2019-10-10
DOC: add python3.7 to spec

mattip:
add missing 3.7 from supported numpy versions


larsoner:
This was done as part of #236, but feel free to push a commit to add 3.8 and reopen if you want

PR #236, opened 2019-10-10
MNT Drop Python 2.7 and 3.4 support

rth:
Closes https://github.com/numpy/numpydoc/issues/235

Should also close https://github.com/numpy/numpydoc/issues/163

For when that issue receives positive feedback. If not now then later.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,3 @@
-from __future__ import division, print_function
-
 import sys
 import os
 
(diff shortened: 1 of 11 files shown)


rth:
The coverage failure seems to be because the total number (of covered) lines decreased.

setup.py
```diff
@@ -36,18 +34,18 @@ def read(fname):
...
                  "Topic :: Documentation",
                  "Programming Language :: Python",
-                 "Programming Language :: Python :: 2",
-                 "Programming Language :: Python :: 2.7",
                  "Programming Language :: Python :: 3",
-                 "Programming Language :: Python :: 3.4",
                  "Programming Language :: Python :: 3.5",
-                 "Programming Language :: Python :: 3.6"],
+                 "Programming Language :: Python :: 3.6",
+                 "Programming Language :: Python :: 3.7",
```
larsoner: We should add 3.8 but can do it later

larsoner:
Thanks @rth !

PR #240, opened 2019-10-25
Adding --validate option __main__ and run new validation

### LINKED ISSUE #213: Validation is not strict enough

datapythonista:
Closes #213

`python -m numpydoc` keeps existing functionality, but when called with `--validate` runs the new validation:

```
(sklearn-dev) [mgarcia@xps numpydoc]$ PYTHONPATH=. python -m numpydoc --validate sklearn.linear_model.LinearRegression
sklearn.linear_model.LinearRegression:GL02:Closing quotes should be placed in the line after the last text in the docstring (do not close the quotes in the same line as the text, or leave a blank line between the last text and the quotes)
sklearn.linear_model.LinearRegression:GL03:Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
sklearn.linear_model.LinearRegression:ES01:No extended summary found
sklearn.linear_model.LinearRegression:PR06:Parameter "fit_intercept" type should use "bool" instead of "boolean"
sklearn.linear_model.LinearRegression:PR08:Parameter "fit_intercept" description should start with a capital letter
sklearn.linear_model.LinearRegression:PR06:Parameter "normalize" type should use "bool" instead of "boolean"
sklearn.linear_model.LinearRegression:PR06:Parameter "copy_X" type should use "bool" instead of "boolean"
sklearn.linear_model.LinearRegression:SA01:See Also section not found
sklearn.linear_model.LinearRegression:EX01:No examples section found
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ script:
   - pytest -v --pyargs numpydoc
   # Making sure the command line options work
   - python -m numpydoc numpydoc.tests.test_main._capture_stdout
-  - ! python -m numpydoc numpydoc.tests.test_main._invalid_docstring
+  - echo '! python -m numpydoc numpydoc.tests.test_main._invalid_docstring' | bash
   - python -m numpydoc --validate numpydoc.tests.test_main._capture_stdout
-  - ! python -m numpydoc --validate numpydoc.tests.test_main._docstring_with_errors
+  - echo '! python -m numpydoc --validate numpydoc.tests.test_main._docstring_with_errors' | bash
   # Build documentation
   - |
     cd ../doc
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,12 @@ script:
     cd dist
     pip install numpydoc* -v
   - pytest -v --pyargs numpydoc
+  # Making sure the command line options work
+  - python -m numpydoc numpydoc.tests.test_main._capture_stdout
+  - ! python -m numpydoc numpydoc.tests.test_main._invalid_docstring
+  - python -m numpydoc --validate numpydoc.tests.test_main._capture_stdout
+  - ! python -m numpydoc --validate numpydoc.tests.test_main._docstring_with_errors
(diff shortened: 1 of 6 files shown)


larsoner:
Sounds good to me. This broke the `test_main.py` test. It would also be good to add a test for this to `.travis.yml`, maybe by calling `numpydoc --validate numpydoc.<something>` and asserting that it works for something good (zero exit code and no output) and for something bad (non-zero exit code and some output)

datapythonista:
Tests are passing now. Not sure exactly what was the goal of mocking modules in the original tests, but I reimplemented all tests in a way that seems simpler to me, and that everything is being tested. I also added calls to `python -m numpydoc` to travis directly to check the status codes.

I don't think the coverage warning is relevant, I think it's being caused by the lack of tests of the `argparse` stuff in `if __name__ == '__main__':`. I'm sure we don't want to test that.

datapythonista:
@larsoner did you have a chance to look at this? I think this should be ready, let me know if it's not.

datapythonista:
Yes, I think ignoring is complex enough to go into a separate PR. We probably want to specify it in `setup.cfg` too, and whether is better to whitelist or blacklist error codes may not be obvious.

larsoner:
Okay let's get this in and keep iterating, thanks @datapythonista

PR #246, opened 2019-11-23
STY: Reword first lines of example.py docstrings

justinludwig:
Running doc/example.py through [pydocstyle](http://www.pydocstyle.org/)
using the "numpy" conventions turns up the following violations for the
module docstring:

* D205: 1 blank line required between summary line and description
  (found 0)
* D400: First line should end with a period (not 'd')
* D404: First word of the docstring should not be `This`
* D415: First line should end with a period, question mark, or
  exclamation point (not 'd')

And the following violation for the `foo()` function docstring:

* D401: First line should be in imperative mood; try rephrasing (found
  'A')

This change slightly rewords the first line of both docstrings to better
model how docstrings should begin:

1. It reformats the module docstring to place the first sentence on its
own line, and remove the superflous "This is the" introduction.
2. It rewords the `foo()` function docstring to rephrase the first line
as a command (following the PEP 257 convention).

After this change, pydocstyle finds no violations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
 before_install:
   - sudo apt-get install texlive texlive-latex-extra latexmk dvipng
   - pip install --upgrade pip setuptools  # Upgrade pip and setuptools to get ones with `wheel` support
-  - pip install pytest pytest-cov numpy matplotlib sphinx${SPHINX_SPEC} codecov
+  - pip install pytest pytest-cov pydocstyle numpy matplotlib sphinx${SPHINX_SPEC} codecov
 script:
   - |
     python setup.py sdist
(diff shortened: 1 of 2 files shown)


larsoner:
@justinludwig it might be useful to add a check to Travis that it complies. Do you want to try adding it here? If not, okay for me to push some commits to your branch?

justinludwig:
>  it might be useful to add a check to Travis that it complies. Do you want to try adding it here?

Sure, I'll try adding that.

justinludwig:
> it might be useful to add a check to Travis that it complies. Do you want to try adding it here?

I added a line to the travis script to run `pydocstyle` on the `example.py` file -- @larsoner  is that what you had in mind?

larsoner:
Yep, eventually we should probably do the whole `numpydoc` module but this is a good start. Thanks @justinludwig

PR #256, opened 2020-04-11
Disable escaping "*" on signature

tk0miya:
On handling autodoc-process-signature event, it's not needed to
escape "*" characters.  The escaping is mainly allowed for
highlighting feature of editors.  So it's not necessary for
generating contents by autodoc.

In addition, since Sphinx-3.0, the escaping is not recommended by
default (see strip_signature_backslash)

refs: https://github.com/sphinx-doc/sphinx/issues/7439

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -564,7 +564,7 @@ class FunctionDoc(NumpyDocString):
             if func is None:
                 raise ValueError("No function or docstring given")
             doc = inspect.getdoc(func) or ''
-        NumpyDocString.__init__(self, doc)
+        NumpyDocString.__init__(self, doc, config)
 
         if not self['Signature'] and func is not None:
             func, func_name = self.get_func()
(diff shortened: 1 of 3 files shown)


larsoner:
Changes LGTM but it would be nice to see the effect before and after. I'd like to see if we can squeak in #258 first, then merge this, if that seems reasonable to you @tk0miya

tk0miya:
>Changes LGTM but it would be nice to see the effect before and after.

Before Sphinx-3.0, backslashes are removed before processing. So this changes nothing.
https://github.com/sphinx-doc/sphinx/blob/2.x/sphinx/directives/__init__.py#L95-L97

Since Sphinx-3.0, backslashes are removed before processing only if `strip_signature_backslash` is enabled (optional). So backslashes will be shown without this change (or enable `strip_signature_backslash` on each project). This is one of incompatible changes in our latest release. Anyway, not to escape also works fine for both versions.
https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/directives/__init__.py#L95-L100

larsoner:
Thanks @tk0miya

PR #258, opened 2020-04-12
ENH: Add args and kwargs to example

larsoner:
I wanted to view the results of #256, but our example does not have `*args, **kwargs` in it. This does that, plus uses circleci-artifacts-redirector-action to make a more direct link to the artifacts.

@rgommers @jnothman can you also enable GitHub actions for this repo? I don't have permissions to do it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,12 @@
+on: [status]
+jobs:
+  circleci_artifacts_redirector_job:
+    runs-on: ubuntu-latest
+    name: Run CircleCI artifacts redirector
+    steps:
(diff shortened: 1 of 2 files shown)


larsoner:
The Travis failure appears to be the one fixed by #256. And you can see that we do indeed have a problem on `master` /  this PR:

https://107-11639513-gh.circle-artifacts.com/0/html/example.html

rgommers:
> @rgommers @jnothman can you also enable GitHub actions for this repo? I don't have permissions to do it.

That already was enabled:

image

Also checked permissions, and bumped `numpydoc-team` from Write to Maintain

PR #259, opened 2020-04-13
Fixes references outside function (#214)

### LINKED ISSUE #214: References outside function
I have seen a similar problem as mentioned [here](https://github.com/numpy/numpydoc/issues/177#issuecomment-487175646). Problem is for docstrings but only for docstrings outside of functions. I have attached an [example](https://github.com/numpy/numpydoc/files/3123855/numpydoctest2.zip), and by running the command `sphinx-build -b singlehtml -d _build/doctrees . _build/singlehtml` l i get the following for output:
56817105-2daf3500-6845-11e9-9571-f758848d6cc7

If I delete `:members:` from manual.rst the docstring outside the function works as expected.
Screenshot 2019-04-27 at 14 21 26

Am i doing anything wrong? I'm running the newest version (0.9.1).

Hoxbro:
Closes #214

### DIFF
--- a/numpydoc/tests/test_full.py
+++ b/numpydoc/tests/test_full.py
@@ -1,4 +1,5 @@
 import os.path as op
+import re
 import shutil
 
 import pytest
@@ -65,3 +66,23 @@ def test_my_function(sphinx_app):
     assert '*args' in html
     # check xref (iterable should link using xref):
     assert 'glossary.html#term-iterable' in html
+
+
+def test_reference(sphinx_app):
+    """Test for bad references e"""
+    out_dir = sphinx_app.outdir
+    html_files = [
+        ["index.html"],
+        ["generated", "numpydoc_test_module.my_function.html"],
+        ["generated", "numpydoc_test_module.MyClass.html"],
+    ]
+
+    for html_file in html_files:
+        html_file = op.join(out_dir, *html_file)
+
+        with open(html_file, 'r') as fid:
+            html = fid.read()
+
+        reference_list = re.findall(r'<a class="fn-backref" href="\#id\d+">(.*)<\/a>', html)
+        for ref in reference_list:
+            assert '-' not in ref  # Bad reference if it contains "-" e.g. R1896e33633d5-1
--- a/numpydoc/tests/test_full.py
+++ b/numpydoc/tests/test_full.py
@@ -69,7 +69,7 @@ def test_my_function(sphinx_app):
 
 
 def test_reference(sphinx_app):
-    """Test for bad references e"""
+    """Test for bad references"""
     out_dir = sphinx_app.outdir
     html_files = [
         ["index.html"],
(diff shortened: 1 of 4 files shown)


larsoner:
@sebp @MatthewFlamm can you see if this fixes #214 for you?

MatthewFlamm:
It still fails on my usage, but I haven't compared precisely to the small test example that @Hoxbro gave in #214.  So there may be still some unresolved issue, or I'm using it incorrectly.  It does work for me on 0.8.0 however.

image

Hoxbro:
@MatthewFlamm Can you give me a small code example of your problem? Then I can look into it.

MatthewFlamm:
These days I unfortunately won't have time to try and simplify my case down to a simple example.  Here is an example from how I used it though:

https://github.com/Merck/rtdpy/blob/master/source/AxialDispersion_cc.rst
https://github.com/Merck/rtdpy/blob/master/rtdpy/ad_cc.py

Hoxbro:
That's okay, the linked files were more than enough. I think that I have fixed your problem to:

image

@larsoner The more underlying problem, seem to be that the DEDUPLICATION_TAG is not necessary placed last in the sibling_section list. I have fixed this with a for-loop but I don´t know if there is a more underlying problem here.

MatthewFlamm:
@Hoxbro Great work!  This works better than 0.8.0 even.  I was getting some duplication of links in 0.8.0, but it was still understandable unlike 0.9+.  Current behavior is below which works in my full code too.

image
image

larsoner:
@Hoxbro do you see some easy way to add a test? For example, you could add something to `numpydoc/tests/tinybuild` (it's a tiny sphinx-buildable thing) that fails to do the right thing on `master`, then assert that there is some suitable substring (or regex) that shows it's working on this PR.

larsoner:
(the tests for `tinybuild` are in `numpydoc/tests/test_full.py`)

numpydoc/tests/test_full.py
```diff
@@ -65,3 +66,23 @@ def test_my_function(sphinx_app):
...
+        ["generated", "numpydoc_test_module.MyClass.html"],
+    ]
+
+    for html_file in html_files:
+        html_file = op.join(out_dir, *html_file)
+
+        with open(html_file, 'r') as fid:
+            html = fid.read()
+
+        reference_list = re.findall(r'<a class="fn-backref" href="\#id\d+">(.*)<\/a>', html)
```
larsoner, line 86:
```suggestion
        reference_list = re.findall(r'<a class="fn-backref" href="\#id\d+">(.*)<\/a>', html)
        assert len(reference_list) == 3
```
Otherwise this could be an empty list and thus not testing what it means to test

Hoxbro: I like it. It is only index.html which have 3 items in the `reference_list`. I have added the check to the test.

larsoner:
Awesome, thanks @Hoxbro !

PR #262, opened 2020-05-07
FIX: Get doc of actual class in test

### LINKED ISSUE #261: Test failure with python 3.9
python-numpydoc fails to build with Python 3.9.0a6.
```
======================================================================
FAIL: test_docscrape.test_see_also_print
----------------------------------------------------------------------
  File "/usr/lib/python3.9/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/builddir/build/BUILD/numpydoc-0.9.2/numpydoc/tests/test_docscrape.py", line 860, in test_see_also_print
    assert(':func:`func_a`, :func:`func_b`' in s)
AssertionError

----------------------------------------------------------------------
Ran 48 tests in 0.354s

FAILED (failures=1)
```
For the build logs, see:
https://copr-be.cloud.fedoraproject.org/results/@python/python3.9/fedora-rawhide-x86_64/01365545-python-numpydoc/

https://bugzilla.redhat.com/show_bug.cgi?id=1830971

larsoner:
Closes #261 

Seems to be caused by https://bugs.python.org/issue40257 / https://docs.python.org/3.9/whatsnew/3.9.html#changes-in-the-python-api. I assume it's actually reasonable / more correct to look at the doc of the class, rather than an instance of the class. @opoplawski can you see if it indeed fixes it for you? Using `deadsnakes` ppa on 20.04 it seems to work for me...

### DIFF
--- a/numpydoc/tests/test_docscrape.py
+++ b/numpydoc/tests/test_docscrape.py
@@ -828,8 +828,7 @@ def test_see_also_print():
         """
         pass
 
-    obj = Dummy()
-    s = str(FunctionDoc(obj, role='func'))
+    s = str(FunctionDoc(Dummy, role='func'))
     assert(':func:`func_a`, :func:`func_b`' in s)
     assert('    some relationship' in s)
     assert(':func:`func_d`' in s)


opoplawski:
Seems to build fine here as well.  Thanks!

larsoner:
Merging so that this can go into 1.0.0

PR #265, opened 2020-05-26
MAINT: Update manifest

larsoner:
This allows `check-manifest` to pass.


larsoner:
I'll go ahead and merge, but if folks see a problem with this I'm happy to open a follow-up PR. I wasn't 100% sure if I should include the `doc/scipy-sphinx-extension`, but I figure in the `sdist` it's nice to have everything in the file required to build the docs, and the source dist tarball is still just 194 kB so it doesn't seem too bad

PR #278, opened 2020-07-10
ENH: Better warning for sections.

Carreau:
1) if the number of -/= is too short/ too long warn, Especially too
short it won't be detected as a section.

2) for duplicate section print the docstring to figure out where the
problem is.


larsoner:
Can you update some tests?

Carreau:
Done.

larsoner:
Thanks @Carreau

PR #280, opened 2020-07-13
BUG: fix role regex.

Carreau:
Spyder sometime uses the full :py:meth: role, this make sure it is
properly handled.


larsoner:
Thanks @Carreau

PR #291, opened 2020-08-04
MAINT: Add test for tuples

larsoner:
Expose a bug with Sphinx 3.x. Hopefully once they push commits and a point release we can remove the `xfail`.


larsoner:
This was fixed in https://github.com/sphinx-doc/sphinx/issues/8052, closing

PR #300, opened 2020-10-31
Add a note to the docstring standard about long 'See Also' entries.

WarrenWeckesser:


rgommers (approved):
Thanks @WarrenWeckesser

PR #308, opened 2021-01-01
BUG: fix an incomplete check in `Reader._error_location`

rgommers:
It's unclear why `self._obj` can be None, but that's what I'm
seeing when trying numpydoc master to build SciPy master.

This fix gives the right traceback.


rgommers:
This is a one-liner fix that cannot break anything, so I'll go ahead and merge it.

PR #309, opened 2021-01-21
DOC: fix erroneous numpydoc_edit_link formatting

sappelhoff:
This is mainly a fix for a missing linebreak in the documentation about `numpydoc_edit_link` which leads to a wrong rendering:

image

see: https://numpydoc.readthedocs.io/en/latest/install.html#sphinx-config-options

I couldn't find which version `numpydoc_edit_link` was deprecated, so if someone knows, it'd be good to add the version after the directive.

Then I also corrected an indent for consistency, this should have no impact on the actual rendering

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -97,7 +97,7 @@ numpydoc_xref_ignore : set or ``"all"``
   ``numpydoc_xref_ignore="all"`` is more convenient than explicitly listing
   terms to ignore in a set.
 numpydoc_edit_link : bool
-  .. deprecated::
+  .. deprecated:: 0.7.0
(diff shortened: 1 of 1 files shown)


sappelhoff:
Tests are failing because the argument to the `.. deprecated::` directive seems to be REQUIRED. Prior to this PR, the first word of the next sentence was parsed as the argument to the directive (erroneously).

`git log -S numpydoc_edit_link` did not help me to figure out in which version this was deprecated :thinking: I'll try to find it via git blame

larsoner:
Straightforward improvement, thanks @sappelhoff !

PR #315, opened 2021-03-21
MAINT: use super() as described by PEP 3135

mwtoews:
This PR refactors `super()` to be simpler, as described by [PEP 3135](https://www.python.org/dev/peps/pep-3135/).

xref https://github.com/numpy/numpy/pull/18648 for more details and background.


rgommers (approved):
Thanks @mwtoews!

PR #325, opened 2021-05-19
MAINT: Python3 classes do not need to inherit from object

mwtoews:
See https://github.com/numpy/numpy/pull/19035


larsoner:
Thanks @mwtoews

PR #327, opened 2021-05-24
improve section detection

### LINKED ISSUE #316: Support (detect?) missing blank line before Returns sections.
It's not the fist time I come across this, but it looks like project regularly forget the blank line between Parameters and Return, leading to the Returns section being merged into Parameters.

That would likely need to update `_read_to_next_section` to use extra heuristics than `read_to_next_empty_line`, 
Only checking for lines with ------, or ====== might not be sufficient as this also would match tracebacks.

https://github.com/Carreau/velin/issues/14 for reference.

keewis:
Potentially fixes #316.

This tries to allow parsing sections which are not separated by blank lines (there should probably be a warning in that case, I'll add that once the general idea has been approved). In order to get this to work used a few tricks (e.g. add a optional `doc` parameter to `_is_at_section` to allow calling it on a different reader) so this might need some refactoring before being truly ready.

cc @Carreau, my main motivation was trying to get `velin` to auto-fix this

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -196,15 +196,15 @@ class NumpyDocString(Mapping):
 
         return doc[i:len(doc)-j]
 
-    def _read_to_next_section(self):
-        section = self._doc.read_to_next_empty_line()
+    def _read_to_next_section(self, doc=None):
+        if doc is None:
+            doc = self._doc
 
-        while not self._is_at_section() and not self._doc.eof():
-            if not self._doc.peek(-1).strip():  # previous line was empty
-                section += ['']
-
-            section += self._doc.read_to_next_empty_line()
-
+        section = []
+        # make sure we actually read to the next section
+        if self._is_at_section(doc=doc):
+            section.append(doc.read())
+        section += doc.read_to_condition(lambda l: self._is_at_section(doc=doc))
         return section
 
     def _read_sections(self):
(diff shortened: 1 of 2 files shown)


keewis:
yielding `StopIteration` seems like a bug: https://github.com/numpy/numpydoc/blob/265ab91fca60192bbc56217d4bae958e326b9b1b/numpydoc/docscrape.py#L214-L219

because it will cause this to fail: https://github.com/numpy/numpydoc/blob/265ab91fca60192bbc56217d4bae958e326b9b1b/numpydoc/docscrape.py#L380-L381

I think this should be fixed, either by changing the yield value to `yield StopIteration, StopIteration` or `yield None, None`, or by raising an error instead (if that's what that was supposed to indicate).

keewis:
Ping. Does anyone have any comments on this?

rgommers:
> Ping. Does anyone have any comments on this?

I think the discussion in gh-316 shows this is not desirable. Detecting in order to raise a better warning or even an exception makes sense though.

keewis:
apologies for pinging and then forgetting about it for a year.

I implemented the requested change, such that it will now warn on every missing empty line (between summary and the first section, or between two sections).

The current implementation comes with a (slight?) performance regression because `NumpyDocString._is_at_section` is now called for every line. I could imagine doing a two-pass implementation, instead: find all separators (multiple `-` or `=` per line) and check if those belong to a section which is not preceded by an empty line, then run the actual extraction.

PR #337, opened 2021-10-25
Correctly validate parameters under the "Other Parameters" section

dcbr:
The "PR01" error is falsely raised when validating doc strings with parameter definitions under the "Other Parameters" section (e.g. for infrequently used parameters). This PR makes sure such parameters are correctly recognized by the validation script (and an extra test is added that previously failed but is now properly supported).

### DIFF
--- a/doc/example.py
+++ b/doc/example.py
@@ -35,7 +35,7 @@ import matplotlib.pyplot as plt
 # numpy module itself, unabbreviated.
 
 
-def foo(var1, var2, *args, long_var_name='hi', **kwargs):
+def foo(var1, var2, *args, long_var_name='hi', only_seldom_used_keyword=0, **kwargs):
     r"""Summarize the function in one line.
 
     Several sentences providing an extended description. Refer to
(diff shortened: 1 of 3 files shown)


dcbr:
So `doc.example.foo` is failing because in that example non-existent parameters are specified in the "Other Parameters" section. The `only_seldom_used_keyword` can be added to the parameter list I guess, as the name clearly suggests the infrequent usage. However I'm unsure what the meaning of the `common_parameters_listed_above` parameter was. Does this mean it is/should be allowed to repeat parameters from the "Parameters" section in the "Other Parameters" section?

larsoner:
> Does this mean it is/should be allowed to repeat parameters from the "Parameters" section in the "Other Parameters" section?

I'm not sure what it means either but I don't think it should be repeated. Your fix of adding `only_seldom_used_keyword` sounds good to me

dcbr:
All right, I had to make the "PR03" test a bit more complicated, as otherwise that example still failed. The new test makes sure the order of the parameters **within each section** is consistent with the order in the signature (whereas previously it was using a concatenation of the parameters in each section).

larsoner:
> (whereas previously it was using a concatenation of the parameters in each section).

I wonder if this part was intentional, in that in theory the infrequently used parameters should come at the end of the args/kwargs list in the signature...

dcbr:
Reverted some of the previous changes, so now once again the signature order has to exactly match the concatenation of the Parameters order and the Other Parameters order.
@larsoner this is ready for review then.

larsoner (approved):

larsoner:
Thanks @dcbr

PR #342, opened 2021-12-16
Avoid mutable default arguments

DimitriPapadopoulos:
While they are not always an actual bug, they are certainly a Python anti-pattern and possible the source of future bugs.

In at least one case, the default mutable argument `config` is actually modified later on by the function.

See for example [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -136,7 +136,7 @@ class NumpyDocString(Mapping):
         'index': {}
     }
 
-    def __init__(self, docstring, config={}):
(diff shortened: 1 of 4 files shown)


larsoner (approved):
Looks straightforward, will merge once CIs run and come back green

larsoner:
Thanks @DimitriPapadopoulos !

PR #345, opened 2021-12-27
Escape newline in docstring.

Carreau:
Otherwise  `'.` is seen as the second parameter.


rgommers (approved):
LGTM, thanks @Carreau

PR #350, opened 2022-01-08
Update deps

jarrodmillman:
sphinx is tested for 1.8+ and some documentation said we support 1.8+
and some said 1.6.5+.  Now it says 1.8+ everywhere.

I also said we need Jinja2 >= 2.10, which was released in 2017.  We
don't test old versions and I am not sure they even work on the versions
of Python we support (3.7 was released in 2018).


jarrodmillman:
Rebased on main.

rgommers (approved):
LGTM, thanks Jarrod

PR #353, opened 2022-01-10
Use fstrings

jarrodmillman:
Ran flynt and visually confirmed that the changes were correct.


rossbar (approved):
LGTM, I didn't notice any bad conversions.

PR #357, opened 2022-01-21
Document release process

jarrodmillman:


jarrodmillman:
https://github.com/networkx/networkx/pull/5274

RELEASE.rst
```diff
@@ -1,6 +1,22 @@
 Release process for ``numpydoc``
 ================================
 
+Introduction
+------------
+
+Example ``__version__``
+
+- 1.8rc1.dev0  # development version of 1.8 release candidate 1
```
rgommers, line 9: These versions are slightly unusual, I'd expect:
```
1.2.0.dev0  # no rc1 in the name
1.2.0rc1
1.2.0
```

rgommers (commented):
Good addition, testing against the largest users is the best way to smoke out potential issues. Overall this LGTM, just one minor comment.

RELEASE.rst
jarrodmillman: That works, but we need a version between `1.2.0rc1` and `1.2.0`.  So we could do:
```
1.2.0.dev0
1.2.0rc1
1.2.0rc2.dev0
1.2.0
```

rgommers: Yes, that's how we do it in numpy and scipy as well

rgommers (approved):
Thanks Jarrod

PR #369, opened 2022-03-29
Upper bound jinja

jarrodmillman:
See https://github.com/sphinx-doc/sphinx/issues/10306.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
 numpy
 matplotlib
-pydata-sphinx-theme
+pydata-sphinx-theme<=0.8.0
 sphinx>2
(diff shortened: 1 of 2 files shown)


rossbar (approved):
See sphinx-doc/sphinx#10291

AIUI this pin will be necessary in order to continue supporting sphinx versions prior to 4.X.

jarrodmillman:
I think it would make sense to drop some of the older sphinx versions for the 1.3 release.

- Sphinx 1.8 was only tested on Python <=3.6 and 3.6 was EOL last year
- Sphinx 2.0 added support for python 3.7.
- Sphinx 3.1 added support for python 3.8, which most of us have as a minimum supported python version per NEP 29.

PR #391, opened 2022-05-08
Use black

jarrodmillman:
I am leaving this draft until it is approved to prevent it from being merged. If it is approved, I will merge it (but not squash it, so I will need to enable that option before merging). The reason I will do that is so I can keep all the black formatting stuff in one commit that I will add to a `.git-blame-ignore-revs` file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,11 @@ repos:
       - id: check-toml
       - id: check-added-large-files
 
+  - repo: https://github.com/psf/black
+    rev: 22.3.0
+    hooks:
+      - id: black
+
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v2.6.2
     hooks:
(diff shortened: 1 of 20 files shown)


larsoner:
I have not kept up to speed on what NumPy has done, but +1 for merge assuming this follows whatever they do, whenever they do it :)

larsoner (approved):
I'll approve but it would be nice if another maintainer (@rgommers ? @jnothman ?) could weigh in as well!

stefanv (approved):

rossbar (approved):
I'm +1 on adding an autoformatter as well. Given the discussion on the numpy mailing list, black seems like the sensible choice. Some of the other concerns for numpy e.g. the formatting for math and large arrays are not really a concern for numpydoc.

larsoner:
Okay sounds like you have the green light to proceed @jarrodmillman !

PR #392, opened 2022-05-10
Test on sphinx 5

jarrodmillman:
Just testing out the beta release. I will update this and mark it as ready once the final 5.0.0 release is made. That should be relatively soon.

See https://github.com/sphinx-doc/sphinx/issues/10312


larsoner:
It might not be a bad idea to have one test run just install with `pip install --upgrade --pre sphinx`, that way when 5.1 beta (or whatever's next) shows up, we'll be testing against it. I've had good experiences in other repos catching problems early that way

jarrodmillman:
See #395.

PR #409, opened 2022-06-08
Require sphinx>4

jarrodmillman:
In #385, we tentatively agreed to drop support for sphinx 3 with numpy 1.4. We are releasing 1.4 a little sooner than planned.

However, the 1.4 release mainly adds support for new versions of docutils. So for folks using sphinx 3, then there is no reason for them to use numpydoc 1.4. Moreover, pydata-sphinx-theme and myst-parser have dropped support for sphinx 3 with their recent releases. Finally, in #393, you can see that numpydoc doesn't support sphinx ~= 3.5 under python 3.10 and above. You can also see in [6108474](https://github.com/numpy/numpydoc/pull/409/commits/61084740a8e1d1ffe65f5231460e33f8ae19f42f) that we require sphinx >= 4.2 under python 3.10 and above (see https://github.com/sphinx-doc/sphinx/issues/9562 for more information).

Maybe it is too aggressive, but I think it may make sense to require sphinx >= 4.2.

- sphinx 4.0.2 released May 20, 2021
- sphinx 4.2 released Sept 12, 2021

If we only require `sphinx>=4.0.2`, how should we document that for Python 3.10+ we require `sphinx>=4.2`?

Thoughts?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,15 +13,19 @@ jobs:
       matrix:
         os: [Ubuntu]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        sphinx-version:
-          ["sphinx==4.0.2", "sphinx==4.5", "sphinx==5.0", "sphinx>=5.0"]
-        exclude:
-          - python-version: "3.10"
-            sphinx-version: "sphinx==4.0.2"
         include:
+          - python-version: "3.7"
+            sphinx-version:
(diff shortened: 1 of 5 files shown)


jarrodmillman:
@larsoner @rossbar Does this look OK? Should we require sphinx >= 4.2, since earlier versions don't work on Python 3.10+?

Alternatively, we could continue supporting 3.0-3.4 for Python < 3.10 (see #410).

PR #410, opened 2022-06-09
Fix CI

jarrodmillman:
Alternative to #409.


jarrodmillman:
This just fixes the CI, so I am just merging it. This will be replaced soon by #409 anyway (or at least after 1.4 is released).

PR #411, opened 2022-06-09
Require sphinx>=4.2

jarrodmillman:
Per discussion in #409, this drops support for sphinx < 4.2. Let's hold off on merging this for a week or so (to see if there are any issues with 1.4).


larsoner:
Great, thanks @jarrodmillman !

PR #415, opened 2022-07-07
ENH: Add support for dict show_inherited_class_members

larsoner:
Over in MNE-Python we subclass some standard types like `list` and `dict` (I know, not a recommended practice) and for these few cases we don't want to show inherited class members. For all other classes, we do.

This PR adds support for `numpydoc_show_inherited_class_members` as a dict allowing mapping on an individual-class-name basis. ~~If people agree this is a reasonable way to go, I'll add a test then it should be good to go.~~

I was hoping just adding `:no-inherited-members:` to a new autosummary template `class_no_inherited_members.rst` and using it for the few classes where I need it would be enough, but I don't see a way to get/see this switch inside NumpyDoc.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -26,10 +26,18 @@ numpydoc_show_class_members : bool
   Whether to show all members of a class in the Methods and Attributes
   sections automatically.
   ``True`` by default.
-numpydoc_show_inherited_class_members : bool
+numpydoc_show_inherited_class_members : bool | dict
   Whether to show all inherited members of a class in the Methods and Attributes
   sections automatically. If it's false, inherited members won't shown.
-  ``True`` by default.
+  ``True`` by default. It can also be a dict mapping names of classes to
+  boolean values (missing keys are treated as ``True``).
+  For example, ``defaultdict(lambda: False, {'mymod.MyClass': True})``
+  would only show inherited class members for ``MyClass``, whereas
+  ``{'mymod.MyClass': False}`` would show inherited class members for all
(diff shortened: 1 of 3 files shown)


larsoner:
For an example in the wild, this is our current docstring with a bunch of `dict` stuff we don't want:

https://mne.tools/dev/generated/mne.Forward.html

Screenshot from 2022-07-07 13-29-54


And this is it with no `dict` stuff:

Screenshot from 2022-07-07 13-29-04

larsoner:
And we actually have a use case where we want to document some subclass methods (from our own mixin classes) but not others (from subclassing dict)... but I'm not sure a good API for this, or how to get autosummary to play nicely with it.

rossbar:
I'll admit I have to look up how autosummary/autodoc work together every time I think about them so I may be thinking about this completely wrong, but is there a way to generate the stubs via `autoclass` (which already has a `no-inherited-members` option) and have `autosummary` pick up those generated stubs instead of generating them itself?

I'm just trying to think of ways to do this with the existing machinery, but I'm no sphinx expert so maybe it isn't possible (or is extra hacky). The approach taken here in conjunction with mne-tools/mne-python#10911 seems like a good way to go about this, so if IYO this is the best way then I think this is reasonable.

larsoner:
I tried messing with autoclass options and Jinja templates for some time but couldn't make it work that way

jarrodmillman (approved):

larsoner:
Okay since there are some approvals I'll go ahead and merge, thanks for looking @jarrodmillman @rossbar !

PR #425, opened 2022-08-17
ENH: Update validate.py to allow parameters with trailing underscores.

stefmolin:
In https://github.com/numpy/numpydoc/issues/62, https://github.com/numpy/numpydoc/pull/144, and https://github.com/numpy/numpydoc/pull/63, changes are discussed for rendering parameters with trailing underscores; however, the `validate.py` logic doesn't seem to be compatible.

Given a docstring like the following the validation logic currently reports PR01 "Parameters {'type_'} not documented" and PR02 "Unknown parameters {'type\\_'}" because it doesn't handle the escape character:
```python
def test_func(type_):
    r"""
    Test the docstring.

    Parameters
    ----------
    type\_ : str
       Some description.
    """
    pass
   ```
 
On the flip side, removing the escape character passes the validation checks, but does not render properly in the docs (renders as `type` instead of `type_`).

### DIFF
--- a/numpydoc/tests/test_validate.py
+++ b/numpydoc/tests/test_validate.py
@@ -1144,6 +1144,7 @@ class TestValidator:
             "other_parameters",
             "warnings",
             "valid_options_in_parameter_description_sets",
+            "parameters_with_trailing_underscores",
         ],
     )
     def test_good_functions(self, capsys, func):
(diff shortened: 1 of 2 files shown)


rossbar (commented):
+1 for this change, but I think the test case might need tweaking, as the added test still passes with the old validation code.

stefmolin:
Sorry about that &ndash; I didn't realize I had to add the function to the call for `test_good_functions()` as well.

rossbar (approved):
> Sorry about that – I didn't realize I had to add the function to the call for test_good_functions() as well.

No worries, I get confused every time I look at the validation test suite :)

This LGTM, thanks @stefmolin !

larsoner:
Thanks @stefmolin !

PR #448, opened 2023-01-24
MAINT: Fix CIs for sphinx 6

larsoner:
Let's see if this fixes CIs -- pin to `< 6` when running `make html`

Also sneaks in a `concurrency` group to save CI resources in case people push repeatedly / quickly


larsoner:
@rossbar feel free to merge if you're happy with this workaround

rossbar (approved):
Thanks @larsoner I agree let's get this in to fix CI

PR #451, opened 2023-02-19
Update blacken-docs repo after owner transfer.

stefmolin:
The `blacken-docs` repo was transferred to a new owner at the end of 2022 ([source](https://github.com/adamchainz/blacken-docs#history)). If you visit the original repository, it now redirects to https://github.com/adamchainz/blacken-docs.


rossbar (approved):
Thanks @stefmolin !

PR #462, opened 2023-06-08
Drop Python 3.7

jarrodmillman:
Python 3.7 security support will be dropped in about two weeks (27 Jun 2023). NEP29 and SPEC0 already recommended projects in the ecosystem drop Python 3.7 support.


larsoner:
Thanks @jarrodmillman !

PR #477, opened 2023-07-03
Fix dynamic version specification.

stefmolin:
I noticed a small bug coming from https://github.com/numpy/numpydoc/pull/474. In order to use the dynamic version, you have to specify where it is coming from otherwise it installs as `numpydoc-0.0.0`. I added the required logic to the `tools/pyproject.toml.in` file and ran the script to update the file.

Before change:

`Successfully installed numpydoc-0.0.0`

After change:

`Successfully installed numpydoc-1.6.0rc1.dev0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,8 @@ packages = [
     'numpydoc',
     'numpydoc.hooks',
 ]
+[tool.setuptools.dynamic.version]
+attr = 'numpydoc.__version__'
 
 [tool.setuptools.package-data]
 numpydoc = [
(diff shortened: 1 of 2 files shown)


jarrodmillman (approved):

stefanv:
If we're compiling the toml anyway, why not inject a static version parsed from the code?

jarrodmillman:
We can make a follow-up PR if we decide to inject a static version parsed from the code.

larsoner:
In several projects I've started using setuptools_scm with good

PR #487, opened 2023-08-17
Update pre-commits

jarrodmillman:


larsoner:
... looks like it worked:

https://results.pre-commit.ci/run/github/11639513/1692286665.X309Hbw3TWq2shr-Hwn_hA

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: f71fa2c1f9cf5cb705f73dffe4b21f7c61470ba9 # frozen: v4.4.0
```
larsoner, line 6: wait... why freeze to specific commits? Shouldn't we let these update and make incremental changes as new bugs etc. are found? Usually the bot will do the heavy lifting for us

jarrodmillman: I just ran `$ pre-commit autoupdate --freeze`. The bot should do it as well, I normally fix the rev and have the bot update monthly to keep down the noise.

larsoner: Okay, feel free to configure the bot as you see fit now that it's enabled!

PR #488, opened 2023-08-17
Add config for pre-commit bot

jarrodmillman:
@larsoner On other projects, I've used this configuration for the pre-commit.ci bot. I am not sure how it works without the config. Thoughts or suggestions?


larsoner:
Thanks @jarrodmillman

larsoner:
> I am not sure how it works without the config.

We use no bot config on MNE-Python and it has worked well so far. People haven't complained about style fixes being pushed to their PRs at least. And the updates don't come super often (1-2 / mo?) but that'll depend on the hooks used. For us it's mostly `black` and `ruff` that update, and they're usually useful.

PR #490, opened 2023-08-17
Update pydata-sphinx-theme

jarrodmillman:
Use more recent sphinx and pydata-sphinx-theme to build documentation. 

I also bumped circleci to use Python 3.11 to build the docs.


pyproject.toml
```diff
@@ -50,8 +50,8 @@ developer = [
 doc = [
     'numpy>=1.22',
     'matplotlib>=3.5',
-    'pydata-sphinx-theme>=0.13',
-    'sphinx>=6',
+    'pydata-sphinx-theme>=0.13.3',
+    'sphinx>=7',
```
larsoner, line 54: this might need a `<7.2` until pydata-sphinx-theme releases a new version, see https://github.com/pydata/pydata-sphinx-theme/issues/1404 (which could have at one point read "incompatibility with Sphinx 7.2")... but they just cut 7.2.1 so we'll see!

larsoner (approved):
Marking for merge-when-green, hopefully it works

larsoner:
we should also get rid of the `style` run, it should be redundant with `pre-commit.ci` now. Feel free to do it here if you want @jarrodmillman, I'm adding a branch protection rule so we can click "merge when green" in the future

pyproject.toml
jarrodmillman: 7.2.1 worked for me on my computer...

PR #493, opened 2023-08-18
TST: Test no-file for source

larsoner:
Not sure this is the best solution but it at least allows things in MNE-Python to pass. Turns out the real issue was us subclassing `list` and that having to associated `file`. But I added another test that failed a different way along the way as well for where we create functions dynamically.

cc @stefmolin @stefanv since you've probably thought about this more than I have! And @jarrodmillman since we should probably get some form of this in for 1.6.

BTW the time to test docstrings in MNE-python went from this before #476:
```
5.15s call     mne/tests/test_docstring_parameters.py::test_docstring_parameters
```
to:
```
49.58s call     mne/tests/test_docstring_parameters.py::test_docstring_parameters
```
So we should indeed prioritize time/efficiency at some point soon :)

### DIFF
--- a/.github/workflows/label-check.yml
+++ b/.github/workflows/label-check.yml
@@ -4,8 +4,10 @@ on:
   pull_request:
     types:
       - opened
+      - repoened
       - labeled
       - unlabeled
+      - synchronize
 
 env:
   LABELS: ${{ join( github.event.pull_request.labels.*.name, ' ' ) }}
--- a/numpydoc/tests/test_validate.py
+++ b/numpydoc/tests/test_validate.py
@@ -41,6 +41,33 @@ def test_get_validation_checks_validity(checks):
         _ = validate.get_validation_checks(checks)
 
 
+class _DummyList(list):
+    """Dummy list class to test validation."""
+
+
+def test_no_file():
+    """Test that validation can be done on functions made on the fly."""
+    # Just a smoke test for now, <list> will have a None filename
+    validate.validate("numpydoc.tests.test_validate._DummyList.clear")
+    # This does something like decorator.FunctionMaker.make
+    src = """\
+def func():
+    '''Do something.'''
+    return 1
+"""
+    evaldict = {}
+    exec(compile(src, "<string>", "single"), evaldict)
+    func = evaldict["func"]
+    func.__source__ = src
+    func.__module__ = "numpydoc.tests.test_validate"
+    assert func() == 1
+    # This should get past the comment reading at least. A properly
+    # wrapped function *would* have source code, too. We could add such
+    # a test later
+    with pytest.raises(OSError, match="could not get source code"):
+        validate.validate(validate.get_doc_object(func))
+
+
 @pytest.mark.parametrize(
     ["file_contents", "expected"],
     [
--- a/numpydoc/tests/test_validate.py
+++ b/numpydoc/tests/test_validate.py
@@ -49,23 +49,6 @@ def test_no_file():
     """Test that validation can be done on functions made on the fly."""
     # Just a smoke test for now, <list> will have a None filename
     validate.validate("numpydoc.tests.test_validate._DummyList.clear")
-    # This does something like decorator.FunctionMaker.make
-    src = """\
-def func():
-    '''Do something.'''
-    return 1
-"""
-    evaldict = {}
-    exec(compile(src, "<string>", "single"), evaldict)
-    func = evaldict["func"]
-    func.__source__ = src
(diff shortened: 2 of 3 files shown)


numpydoc/validate.py
```diff
@@ -125,8 +127,12 @@ def extract_ignore_validation_comments(filepath: os.PathLike) -> Dict[int, List[
     dict[int, list[str]]
         Mapping of line number to a list of checks to ignore.
     """
-    with open(filepath) as file:
```
stefmolin, line 128: I would tweak this slightly.

```python
numpydoc_ignore_comments = {}

try:
    with open(filepath) as file:
        last_declaration = 1
        declarations = ["def", "class"]
        for token in tokenize.generate_tokens(file.readline):
            if token.type == tokenize.NAME and token.string in declarations:
                last_declaration = token.start[0]
            if token.type == tokenize.COMMENT:
                match = re.match(IGNORE_COMMENT_PATTERN, token.string)
                if match:
                    rules = match.group(1).split(",")
                    numpydoc_ignore_comments[last_declaration] = rules
except (OSError, TypeError):  # can be None, nonexistent, or unreadable
    pass

return numpydoc_ignore_comments
```

numpydoc/tests/test_validate.py
```diff
@@ -41,6 +41,33 @@ def test_get_validation_checks_validity(checks):
...
+    evaldict = {}
+    exec(compile(src, "<string>", "single"), evaldict)
+    func = evaldict["func"]
+    func.__source__ = src
+    func.__module__ = "numpydoc.tests.test_validate"
+    assert func() == 1
+    # This should get past the comment reading at least. A properly
+    # wrapped function *would* have source code, too. We could add such
+    # a test later
+    with pytest.raises(OSError, match="could not get source code"):
```
stefmolin, line 67: Bringing it back to my [earlier question](https://github.com/numpy/numpydoc/pull/476#discussion_r1297785873), this would mean that validation wasn't run in these cases because it would have failed during the ast parsing [here](https://github.com/numpy/numpydoc/blob/main/numpydoc/validate.py#L485)  (4 year old code), and therefore not returned any findings. In this case, I wonder if it would make more sense to be explicit why this happens in an error message or support reporting on a subset of the checks that don't require this.

numpydoc/validate.py
larsoner: I actually prefer the current approach as I always try to limit the scope of `try/except` as much as possible. With your code there is a single return path, which is nice. However the scoping for the `try/except` is broader, which means you risk accidentally catching a TypeError somewhere in the remaining 10 lines that do more than just "try to open the file". And I don't think the interpretation of the current code is too bad: "try to open the file and if you can't, return an empty set. if you can, proceed to parse it"

numpydoc/tests/test_validate.py
larsoner: Maybe but I want to stress this is really a corner case I only hit while trying to find the *real motivating* case above. So actually I think it's better for now if I just remove this so we aren't confused by it...

larsoner:
I also realized that `extract_ignore_validation_comments` was being called for every `:py:obj:` to be validated, and that this is probably a big overhead when you only need to do it once per *source file* (as each of those could have tens/hundreds/thousands(?) of docstrings in them). Based on this:
```
$ find mne-python/mne -name "*.py" | wc -l
609
$ find matplotlib/lib/ -name "*.py" | wc -l
290
$ find pandas/pandas/ -name "*.py" | wc -l
1376
$ find scipy/scipy/ -name "*.py" | wc -l
924
$ find numpy/numpy/ -name "*.py" | wc -l
495
```
I added a `@functools.lru_cache(maxsize=2000)` to `extract_ignore_validation_comments`. This dropped the time from ~50s to ~10s for MNE-Python, which seems well worth the slight memory usage increase required to store the output dicts.

.github/workflows/label-check.yml
```diff
@@ -4,8 +4,10 @@ on:
   pull_request:
     types:
       - opened
+      - repoened
       - labeled
       - unlabeled
+      - synchronize
```
larsoner, line 10: @jarrodmillman note I added this to the event for the label check -- otherwise it only runs on the first commit and no subsequent ones. I think we want it to run for all so we can add it to the requirements for merge-when-green

larsoner: (you can see it ran on 12e59fd where I added this to the event list but not on the preceding commit f18e780)

jarrodmillman (approved):

jarrodmillman:
Thanks!

Issue #494, opened 2023-08-19
Numpydoc validation - ignore setters


akaszynski:
Excellent work on implementing the numpydoc validation as a pre-commit hook.

I've noticed that it is flagging getters and sellers. For example:

```py
    @property
    def x(self) -> np.ndarray:
        """Return or set the coordinates along the X-direction.

        Examples
        --------
        Return the x coordinates of a RectilinearGrid.

        >>> import numpy as np
        >>> import pyvista
        >>> xrng = np.arange(-10, 10, 10, dtype=float)
        >>> yrng = np.arange(-10, 10, 10, dtype=float)
        >>> zrng = np.arange(-10, 10, 10, dtype=float)
        >>> grid = pyvista.RectilinearGrid(xrng, yrng, zrng)
        >>> grid.x
        array([-10.,   0.])

        Set the x coordinates of a RectilinearGrid.

        >>> grid.x = [-10.0, 0.0, 10.0]
        >>> grid.x
        array([-10.,   0.,  10.])

        """
        return convert_array(self.GetXCoordinates())

    @x.setter
    def x(self, coords: Sequence):
        self.SetXCoordinates(convert_array(coords))
        self._update_dimensions()
        self.Modified()
```

Is it possible to ignore the returns (RT01) of properties and the docstring all together of setters (GL08)?

Issue #486, opened 2023-08-08
How to document properties parameters?


sdiebolt:
I'm wondering how to porperly document property getters and setters. Until now, I've been documenting everything in the getter, leaving the setter without docstring. That way, the generated HTML documentation shows the whole property documentation. However, I get a `PR02: Unknown parameters` warning on my getters since they have a `Parameters` section documenting the parameters of the setter. What am I doing wrong?

timhoffm:
Why do you need a parameters section on properties at all? Properties are logically attributes, i.e. repesent one value that can be read or written. Docstrings are for the user and for them properties appear attribute-like. The functions and there parameters are an implementation detail that's not relevant for the documentation.

I would suggest to document it like this:

```
class Rectangle:
    ...

    @property
    def width(self) -> float:
        """The width of the rectangle."""
        return self._width

    @width.setter
    def width(self, w: float) -> None:
        self._width = w
```

This is analogous to the attribute definition

```
class Rectangle:

    width : float
    """The width of the rectangle."""
```

PR #496, opened 2023-08-24
WIP, BUG: validator now handles decorators

### LINKED ISSUE #495: `numpydoc ignore` inline comment not recognized when using decorators
When using decorators, the `numpydoc ignore` inline comment does not seem to be recognized by either `numpydoc --validate` or the validation during Sphinx build. Here's a reprex, stored as `test.py`:

```python
import functools


def test_decorator(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        return func(*args, **kwargs)

    return wrapper


@test_decorator
def test_function():  # numpydoc ignore=PR02
    """Test.

    Parameters
    ----------
    x : int
        Test.
    """
    print("test")
```

Using `python -m numpydoc --validate test.test_function` raises the warning `test.test_function:PR02:Unknown parameters {'x'}` when the decorator is used, but doesn't when the decorator is removed. Same thing when using validation during a Sphinx build. Interestingly, validation using `python -m numpydoc.hooks.validate_docstrings test.py` does not raise the PR02 warning, even when using the decorator.

sdiebolt:
This PR fixes a bug that prevented `Validator.source_file_def_line` to return the correct line number of the `def` keyword when the function used decorators.

This issue made `validate` miss inline ignore comments: `extract_ignore_validation_comments` would return a dictionary with keys corresponding to the line numbers of `def` and `class` keywords, but it would then be indexed `Validator.source_file_def_line`, which would correspond to the line of the first decorator, not the `def`/`class` keyword.

I'll try to add a regression test ASAP. 

Closes #495.


numpydoc/validate.py
```diff
@@ -295,7 +295,17 @@ def source_file_def_line(self):
...
         """
         try:
-            return inspect.getsourcelines(self.code_obj)[-1]
+            sourcelines = inspect.getsourcelines(self.code_obj)
+            # getsourcelines will return the line of the first decorator found for the
+            # current function. We have to find the def declaration after that.
+            def_lines = [
+                i
+                for i, x in enumerate(
+                    [re.match("^ *(def|class)", s) for s in sourcelines[0]]
```
larsoner, line 304: I would make this tiny tweak but I doubt it matters in practice
```suggestion
                    [re.match("^ *(def|class) ", s) for s in sourcelines[0]]
```

larsoner (approved):
Looks good so far, let me know when the test is there and things are green and I think we can merge!

sdiebolt:
@larsoner I've written a simple regression test, but I'm unsure if this fits in the existing testing pipeline. Please don't hesitate to indicate any issue or possible improvement, I don't mind rewritting these to better fit the existing tests.

On another note, I've just noticed that a similar bug still exists for `@property` decorated methods. The fix proposed here does not work, as `inspect.getsourcelines` raises a `TypeError` when passed a property object, leading `Validator.source_file_def_line` to return `None`. This may be related to #486 and #494. 

I think we may be able to fix this by passing the `fget` attribute of property objects, which is the getter function, to `inspect.getsourcelines`. Shall I fix this in this PR or open a new one?

larsoner:
Failure looks related, can you check?

> I think we may be able to fix this by passing the fget attribute of property objects, which is the getter function, to inspect.getsourcelines. I think numpydoc only parses the getter anyway, so that should not be a problem. Shall I fix this in this PR or open a new one?

Probably better as a follow-up PR if we can merge this one quickly, which I think we can. But the current changes are small so if you think it makes sense to combine them then I can live with that, too.

sdiebolt:
Ok this is a weird one. Seems like `inspect.getsourcelines` does not catch decorators from a decorated class in Python 3.8. It does for function though. I'm not sure why this is the case. I'll modify the tests to take it into account.

sdiebolt:
Feel free to merge if everything is OK and I'll start the follow up PR. :slightly_smiling_face:

larsoner:
Marking for merge-when-green, thanks in advance @sdiebolt !

PR #499, opened 2023-08-25
Fix clean_backref for extensions that have backrefs to inline elements.

mcmtroffaes:
Originally reported here:

https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/277

### DIFF
--- a/numpydoc/numpydoc.py
+++ b/numpydoc/numpydoc.py
@@ -24,7 +24,7 @@ from collections.abc import Callable
 import hashlib
 import itertools
 
-from docutils.nodes import citation, Text, section, comment, reference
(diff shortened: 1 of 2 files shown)


larsoner (commented):
Seems reasonable to me. Ideally it would be nice if we could add a test that fails on `main` and passes on this PR -- @mcmtroffaes how difficult do you think this would be?

mcmtroffaes:
Thanks for the quick response. Yes, I think I should be able to add a test.

mcmtroffaes:
I've added a test.

rossbar (approved):
LGTM, thanks @mcmtroffaes !

mcmtroffaes:
Awesome, many thanks for merging!

PR #500, opened 2023-08-30
BUG: validator now handles properties

### LINKED ISSUE #495: `numpydoc ignore` inline comment not recognized when using decorators

sdiebolt:
This PR fixes a bug that prevented `Validator.source_file_def_line` and `Validator.source_file_name` to return correct values on `@property` and `@functools.cached_property` decorated methods.

Similarly to #496, this issue made `validate` miss inline ignore comments on properties, since the definition line and source file name of properties would be `None`.

Closes #495.

### DIFF
--- a/numpydoc/tests/test_validate.py
+++ b/numpydoc/tests/test_validate.py
@@ -1,7 +1,8 @@
 import pytest
 import sys
 import warnings
-from inspect import getsourcelines
+from functools import cached_property
+from inspect import getsourcelines, getsourcefile
 
 from numpydoc import validate
 import numpydoc.tests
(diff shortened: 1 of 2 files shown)


larsoner (approved):

larsoner:
Thanks @sdiebolt !

PR #524, opened 2024-01-09
MAINT: Replace NameConstant with Constant

larsoner:
Let's see if it works!


cmarmo:
Seems like it's working... I'm going to use this one to patch the debian packaging... :grin:

larsoner:
This seems like a trivial change suggested by the Python docs about the deprecation so I'll self-merge

PR #525, opened 2024-01-14
DOC: Clarify recommendations regarding use of backticks

mdhaber:
The Style guide currently gives the following recommendations about backticks:

In the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) section:

> Enclose variables in single backticks

In [Other points to keep in mind](https://numpydoc.readthedocs.io/en/latest/format.html#other-points-to-keep-in-mind):

> For inline display use double backticks

In [Common reST concepts](https://numpydoc.readthedocs.io/en/latest/format.html#common-rest-concepts):

> Variable, module, function, and class names should be written between single back-ticks (`numpy`).

Contributors (and even maintainers) frequently interpret and/or apply these recommendations inconsistently. This PR attempts to clarify the recommendations to avoid multiple interpretations.

Related issues: 
- https://github.com/sympy/sympy/issues/13519

Follow-up work includes:
- scientific-python/summit-2024#23
- scientific-python/summit-2024#24
- numpy/numpydoc#484


doc/format.rst
```diff
@@ -729,17 +740,27 @@ Other points to keep in mind
...
+
+- Module, function, and class names should be enclosed within ```single
+  backticks```. These are intended to render as hyperlinks (e.g. `numpy`). If
+  the hyperlinks do not render as intended, [insert reference to documentation
+  here].
+- References to parameters, attributes, and methods defined within the same
+  docstring should be enclosed within ```single backticks```. These are
+  currently rendered in *italics*, but there are plans to render a
+  ``monospaced`` hyperlink to the relevant definition in a future version of
+  numpydoc.
```
mdhaber, line 761: Perhaps this should be removed, but I wanted to mention it because it is the only reason I can see for using single backticks instead of double backticks. Otherwise, use of single backticks only causes problems with name collisions and renders in italics (which usually means "emphasis") where monospace (which means code) would be more clear.

doc/format.rst
```diff
@@ -175,8 +175,9 @@ respective types.
   y
       Description of parameter `y` (with type not specified).
 
-Enclose variables in single backticks.  The colon must be preceded
-by a space, or omitted if the type is absent.
+The colon must be preceded by a space, or omitted if the type is absent.
+When referring to a parameter in the description field or elsewhere within
```
stefanv, line 179: Perhaps this would be simpler as "referring to any parameter of the function being documented", or something along that vein?

doc/format.rst
```diff
@@ -549,6 +550,8 @@ not explicitly imported, `.. plot::` can be used directly if
 Documenting classes
 -------------------
 
+.. _classdoc:
```
stefanv, line 553: Is this a new / standard development?  I'm not familiar with it.

doc/format.rst
```diff
@@ -717,7 +726,9 @@ Other points to keep in mind
   (i.e. scalar types, sequence types), those arguments can be documented
   with type `array_like`.
 
-* Links : If you need to include hyperlinks in your docstring, note that
+* Links : Sphinx will automatically create hyperlinks to module, function,
```
stefanv, line 729: Is this recommended practice? I think most libraries use `:pyfunc:` etc. Genuinely asking, I don't know.

stefanv (commented):
Thanks @mdhaber! This looks good overall; I just have a few naive questions.

doc/format.rst
mdhaber: I think it defines a label so I can refer to this section in the "Other points to keep in mind" section below. I'll have to check whether it rendered correctly.

doc/format.rst
mdhaber: It's what is done in SciPy, I think. I think Sphinx only needs more information if there is a name collision.

doc/format.rst
mdhaber: So basically remove "in the description field or...". That's probably fine. I think I wanted to point out specifically that this advice was not limited to references within the parameter documentation, but maybe it's better overall to just keep it simple.

doc/format.rst
```diff
@@ -717,7 +726,9 @@ Other points to keep in mind
...
+  and class documentation if a recognized name is included within single
+  backticks (e.g. `numpy`). If you need to include other hyperlinks, note that
```
mdhaber, line 731:
```suggestion
  backticks (e.g. ```numpy```). If you need to include other hyperlinks, note that
```

rossbar (commented):
Thanks for bringing this up @mdhaber ! There's a lot of nice improvements here, though I think a lot of the recommendations on what to enclose in single backticks are beyond numpydoc's purview and shouldn't necessarily be recommended here.

The linking via single backticks is handled by sphinx proper, not numpydoc, and depends on the configuration and thoroughness of auto-generated stubs for each individual project.

I tend to think of the markup as:
 - double-backticks == monospace formatting with no linking
 - single-backticks == attempted linking

The thing that breaks the above is numpydoc's parameter recommendation, which is what I'd like to fix in #484

doc/format.rst
```diff
@@ -562,10 +565,12 @@ section, may be used to describe non-method attributes of the class::
   Attributes
   ----------
   x : float
-      The X coordinate.
+      Description of attribute `x`.
   y : float
-      The Y coordinate.
+      Description of attribute `y`.
 
+When referring to an attribute in the description field or elsewhere within
```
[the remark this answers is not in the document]
mdhaber: I see. I would prefer not to remove these entirely. 

Is it within numpydoc's purview to recommend that these be rendered as links?

If so, do I understand correctly that typical options will be *just*, single backticks, single backticks preceded by `:attr:`, or single backticks preceded by some other role, and that the choice is up to the project?

doc/format.rst
stefanv: Where else can the variable live? Returns? Class docs, etc.?

doc/format.rst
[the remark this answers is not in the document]
stefanv: The single-backquote-around-a-parameter concept is part of the original numpydoc specification, though, so I'm not sure we can steer clear of it.

Not sure what to say about projects that reconfigure the default role. Technically, their docstrings are no longer numpydoc compliant, but it's not a big deal as long as they consistently use whichever roles they do.

doc/format.rst
mdhaber, line 179:
```suggestion
When referring to a parameter elsewhere within
```

doc/format.rst
```diff
@@ -175,8 +175,9 @@ respective types.
   y
       Description of parameter `y` (with type not specified).
 
-Enclose variables in single backticks.  The colon must be preceded
-by a space, or omitted if the type is absent.
+The colon must be preceded by a space, or omitted if the type is absent.
+When referring to a parameter in the description field or elsewhere within
+the same function or class docstring, enclose its name in single backticks.
```
mdhaber, line 180:
```suggestion
the docstring, enclose its name in single backticks.
```

doc/format.rst
mdhaber: I'm not sure I understand. Is the question about what constitutes a "parameter", where else a parameter might be mentioned, or something else?

doc/format.rst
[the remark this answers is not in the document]
mdhaber: Thanks @stefanv for bringing this up again.

@rossbar can you suggest a compromise here? I think we need to address methods and attributes in some capacity. What *can* we say about them?

doc/format.rst
```diff
@@ -606,6 +611,8 @@ becomes useful to have an additional **Methods** section:
 
       """
 
+When referring to a method in the description field or elsewhere within
```
mdhaber, line 614:
```suggestion
When referring to a method elsewhere within
```

doc/format.rst
mdhaber: It's fine with me to add something about this being a project specific thing and mention the more explicit options, but if it is very common for projects to use single backticks, I think something should be mentioned. What can we say without over-stepping numpydoc's role?

doc/format.rst
```diff
@@ -729,17 +740,27 @@ Other points to keep in mind
...
-Variable, module, function, and class names should be written between
-single back-ticks (```numpy```).
+Use ``*italics*``, ``**bold**`` if needed in any explanations.
+
+Use of backticks in reST is a common point of confusion it is different from
+markdown:
+
+- Module, function, and class names should be enclosed within ```single
+  backticks```. These are intended to render as hyperlinks (e.g. `numpy`). If
+  the hyperlinks do not render as intended, [insert reference to documentation
```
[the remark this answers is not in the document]
mdhaber: Suggestions on that to include here?

mdhaber (commented):
Thanks all. @rossbar can you make suggestions on what we can write in these tricky spots? I'd rather not remove them altogether. The intent here was to make numpydoc's recommendations about backticks usage something that can be read and followed. If numpydoc doesn't have a requirement about something, maybe that's fine, but given all the misuse and confusion (I can produce dozens of PRs if need be), I think it's worth saying *something*.

doc/format.rst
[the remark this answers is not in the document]
rossbar: > The single-backquote-around-a-parameter concept is part of the original numpydoc specification, though, so I'm not sure we can steer clear of it.

Yes, but this is only for the function parameters - recommending that class attributes and methods also be wrapped in single backticks relates to sphinx, not numpydoc, and is nowhere currently mentioned in the numpydoc standard. Generally speaking it's a fine thing to recommend, and if this were a sphinx tutorial I would be +1 for adding it - I'm only leery because this document "defines" the numpydoc docstring standard, and as written this would likely introduce more broken links in documentation (nor does it relate to the "function parameters should be in single backticks" recommendation that has been around since the beginning).

> Not sure what to say about projects that reconfigure the default role. Technically, their docstrings are no longer numpydoc compliant

That's not strictly true - the numpydoc standard doesn't prescribe how projects configure the `default_role` and there are certainly cases where projects may want to override it. To make matters even more confusing, the `default_role` configuration value [can be overridden with the `.. default-role::` directive on a per-file basis](https://docutils.sourceforge.io/docs/ref/rst/directives.html#default-role). The *only* place where there's currently a collision is the long-standing "function parameters should be in single backticks" recommendation. Currently, numpydoc recommends this but does not implement anything specific for it, so the rendered docs end up using the `default_role`. The real solution here is to fix this in numpydoc - see #484.

> can you suggest a compromise here? I think we need to address methods and attributes in some capacity. What can we say about them?

Since this is generally good advice, I'd be perfectly happy to put it in with softer wording; i.e. something that is very clearly a suggestion and not a necessary part of docstring formatting. Perhaps a less wordy version of:

```
When combined with `sphinx.ext.autodoc` or `sphinx.ext.autosummary`, included
{attributes/methods} become valid link targets and can be referenced elsewhere in the
docstring with single backticks. See <link to autoclass> for details.
```

This is hopefully clear, concise, and should do the job for ~95% of use-cases in practice. I'd personally be inclined to include a footnote after "single backticks" to add an additional note about the default role, e.g.

```
docstring with single backticks[^1]. See <link to autoclass> for details.

[^1]: Depending on the configuration of the `default_role <link to default role>`_ for your project.
```

 but it's not necessary.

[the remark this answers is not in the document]
stefanv: :+1: My expectation is only for backticks, anywhere in the docstring, to link to parameter names. It looks like we are in agreement on that, and I'm fine with whatever else is added as Sphinx guidance.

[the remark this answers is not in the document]
mdhaber: > "function parameters should be in single backticks"

Is that documented, though? I see "Enclose _variables_ in single backticks" (emphasis mine). That is in the Parameters section, but I thought that if "variables" were limited to "function parameters" it would say so. The ambiguity there is what led me to try to clarify.

> `See <link to autoclass> for details.`
Is this what you have in mind?
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoclass

> `Depending on the configuration of the `default_role <link to default role>`_ for your project.`
This?
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role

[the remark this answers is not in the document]
stefanv: My best guess is that the intent was to state "function parameters", but 2008 is a long time ago.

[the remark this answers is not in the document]
rossbar: > Why does numpydoc make a recommendation about single backticks around parameters if it doesn't do anything with them?

Yeah this is likely an accident of fate - single backticks has always changed the formatting in the rendered docs (even if it was only to the formatting of broken rst links)

doc/format.rst
rossbar: I'd just remove this - it's probably more detail than most users are interested in (and is not necessarily correct - the actual rendering of e.g. italics is actually dictated by the sphinx theme).

doc/format.rst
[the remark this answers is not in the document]
rossbar: To me, the important distinction to make is: double-backticks is for monospace/code literals, single backticks are for references. If you want to attempt to link to something, use single backticks. If you just want monospace formatting without an attempted link, use double-backticks (the exception of course being numpydoc parameters).

rossbar (commented):
FWIW I found this summary from @jnothman over in a related scikit-learn issue very relevant: https://github.com/scikit-learn/scikit-learn/issues/11186#issuecomment-394279230

timhoffm:
IMHO, the recommendation for single backticks should be complemented with an explanation

> Single backticks are historically used as a means of highlighting. Sphinx tries to resolve them with the `default_role`, which in the default configuration results in italics. If you use a different `default_role` in Sphinx you may need to resort to different ways of highlighting, e.g. double backticks or `*param*` to prevent unintended links or link errors.

doc/format.rst
```diff
@@ -175,8 +175,9 @@ respective types.
   y
       Description of parameter `y` (with type not specified).
 
-Enclose variables in single backticks.  The colon must be preceded
-by a space, or omitted if the type is absent.
+The colon must be preceded by a space, or omitted if the type is absent.
+When referring to a parameter anywhere within the docstring, enclose its
+name in single backticks.
```
Carreau, line 180: :+1:

On a separate PR I hope we can discuss making colon mandatory in the long term for parsing simplicity.

doc/format.rst
```diff
@@ -690,11 +697,11 @@ belong in docstrings.
...
 * Equations : as discussed in the :ref:`Notes <notes>` section above, LaTeX
-  formatting should be kept to a minimum.  Often it's possible to show equations as
-  Python code or pseudo-code instead, which is much more readable in a
-  terminal.  For inline display use double backticks (like ``y = np.sin(x)``).
-  For display with blank lines above and below, use a double colon and indent
-  the code, like::
+  formatting should be kept to a minimum.  Often it's possible to show
+  equations as Python code or pseudo-code instead, which is much more readable
+  in a terminal.  For inline display of code use double backticks,
+  like ````y = np.sin(x)````. For display with blank lines above and below,
```
Carreau, line 703: I think you need a space between each pair of double backticks, or maybe backslashes, that does not render properly on github rendered view for me.

Carreau: Nevermind its an issue with the github renderer.

doc/format.rst
```diff
@@ -562,10 +565,12 @@ section, may be used to describe non-method attributes of the class::
...
   ----------
   x : float
-      The X coordinate.
+      Description of attribute `x`.
   y : float
-      The Y coordinate.
+      Description of attribute `y`.
 
+When referring to an attribute anywhere within the docstring, enclose its
+name in single backticks.
```
mdhaber, line 573: Let's add a footnote about this depending on configuration of default role.

doc/format.rst
mdhaber: Let's add the common footnote about default role.

doc/format.rst
```diff
@@ -717,9 +724,12 @@ Other points to keep in mind
   (i.e. scalar types, sequence types), those arguments can be documented
   with type `array_like`.
 
-* Links : If you need to include hyperlinks in your docstring, note that
-  some docstring sections are not parsed as standard reST, and in these
-  sections, numpydoc may become confused by hyperlink targets such as::
+* Links : Sphinx will automatically create hyperlinks to module, function,
```
mdhaber, line 727: Don't mention Sphinx

doc/format.rst
```diff
@@ -729,17 +739,31 @@ Other points to keep in mind
...
+- Module, function, and class names should render as hyperlinks in monospaced
+  font (e.g. `numpy`), and depending on project settings, it may be sufficient
+  to enclose them in single backticks. If the hyperlink does not render as
+  intended, explicitly include the appropriate role and/or namespace.
+- References to parameters, attributes, and methods defined within the same
+  docstring should also render as hyperlinks in monospaced font. Unless the
+  default role is reconfigured, the recommendation is to enclosed these within
+  ```single backticks```. These typically render as broken links (commonly
+  in *italics*), but there are plans to render a ``monospaced`` hyperlink in
+  a future version of numpydoc.
```
mdhaber, line 764:
```suggestion
  ```single backticks```. These typically render as broken links, but there are plans to render a ``monospaced`` hyperlink in
  a future version of numpydoc.
```

doc/format.rst
mdhaber: @rossbar will add this in a separate PR.

rossbar (approved):
Awesome, thanks for all the hard work here @mdhaber . I think the current wording recommendations are spot on. We can follow-up with more detailed "troubleshooting" docs to handle the cases when single-backtick behave differently than what users expect.

mdhaber:
@rossbar @Carreau @stefanv something not covered by this PR or pydata/pydata-sphinx-theme#1852 is whether short literals like `True`, `None`, and `0` should appear in monospaced font when they refer to code (e.g. capitalized within a sentence). For the sake of consistency, I'd like for the style guide to recommend that they be rendered as code (and that means double backticks).

stefanv:
That is probably more consistent, but would require rewriting nearly all docstrings.
/cc @lagru who is working on [docstub](https://github.com/scientific-python/docstub)

mdhaber:
Compliance with this style guide is always optional. The intent is to reduce future inconsistency.

lagru:
As far as I understand this, this should have no impact on docstub.

Carreau:
I don't see any reason why `None`, `True` and other constant would not be single backticks as they should be crossrefed by intersphinx (https://docs.python.org/3/library/constants.html#True).

Imho the discussion of wether it should use single or double backticks should only be a semantic one, not a visual style one. 

In all case I think we can ask: 
 - Can it be replaced by a directive without changes to the content of the backtick?
     - Yes -> Single
     - No -> Double. 

For example `` `True` `` -> `` :py:obj:`True` ``, yep single. `` `from numpy import sin` `` -> :?:??:`from numpy import sin` `` I don't see what we can use -> double.

Carreau:
I don't see any reason why `None`, `True` and other constant would not be single backticks as they should be crossrefed by intersphinx (https://docs.python.org/3/library/constants.html#True).

Imho the discussion of wether it should use single or double backticks should only be a semantic one, not a visual style one. 

In all case I think we can ask: 
 - Can it be replaced by a directive without changes to the content of the backtick?
     - Yes -> Single
     - No -> Double. 

For example `` `True` `` -> `` :py:obj:`True` ``, yep single. `` `from numpy import sin` `` -> :?:??:`from numpy import sin` `` I don't see what we can use -> double.

mdhaber:
Ah, single is even better then. I didn't consider that they would link.

Carreau:
see #565 for my take on this.

Issue #257, opened 2020-04-11
Overly verbose output for namedtuples


anntzer:
numpydoc generates rather verbose outputs for namedtuples, especially when combined with autosummary:
```
$ sphinx-quickstart -q -m --sep -pfoo -afoo --extensions=sphinx.ext.autosummary,numpydoc
```
then edit `source/index.rst` to contain
```rst
.. autosummary::
   foo.Foo

.. automodule:: foo
   :members:
```
and `foo.py` to contain
```python
from collections import namedtuple

Foo = namedtuple("Foo", "bar baz")
```
then build with
```
PYTHONPATH=. make html
```
This results in
![bad](https://user-images.githubusercontent.com/1322974/79051653-94358a80-7c31-11ea-8925-fbf9a4bb5ff0.png)

Compare with the result using either not numpydoc at all, or sphinx.ext.napoleon instead:
![good](https://user-images.githubusercontent.com/1322974/79051660-a6172d80-7c31-11ea-85d2-f8dcdabfdc3c.png)

I guess whether "count" and "index" should be listed as methods can be argued over, but having "bar" and "baz" being listed twice seems clearly too much, and the "Attributes" entry in the autosummary table is not so great either.

pitrou:
Also, if one tries to actually document the fields, e.g.:
```python

_Foo = namedtuple("_Foo", "bar")

class Foo(_Foo):
    """
    A Foo.

    Attributes
    ----------
    bar : str
        The bar contents of Foo.
    """
    __slots__ = ()
```

then both the documented description ("The bar contents of Foo") _and_ the docstring autogenerated by namedtuple ("Alias for field number 0") are output by numpydoc.

fohrloop:
I think this is fairly easy to fix. A quick and dirty modification in `numpydoc/docscrape.py::ClassDoc` does it

Change `ClassDoc.properties`  to

```python 
    @property
    def properties(self):
        if self._cls is None:
            return []
        return [
            name
            for name, func in inspect.getmembers(self._cls)
            if (
                not name.startswith("_")
                and not self._should_skip_member(name, self._cls) #this added 
                and (
                    func is None
                    or isinstance(func, (property, cached_property))
                    or inspect.isdatadescriptor(func)
                )
                and self._is_show_member(name)
            )
        ]
```

and add `ClassDoc._should_skip_member`:

```python
    @staticmethod
    def _should_skip_member(name, klass):
[...]

PR #527, opened 2024-02-14
fix 'Alias for field number X' problem with NamedTuples

fohrloop:
Fixes https://github.com/numpy/numpydoc/issues/257

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -706,6 +706,7 @@ class ClassDoc(NumpyDocString):
             for name, func in inspect.getmembers(self._cls)
             if (
                 not name.startswith("_")
+                and not self._should_skip_member(name, self._cls)
                 and (
                     func is None
                     or isinstance(func, (property, cached_property))
@@ -714,6 +715,19 @@ class ClassDoc(NumpyDocString):
                 and self._is_show_member(name)
             )
         ]
+
+    @staticmethod
+    def _should_skip_member(name, klass):
+        if (
+            # Namedtuples should skip everything in their ._fields as the
+            # docstrings for each of the members is: "Alias for field number X"
+            issubclass(klass, tuple)
+            and hasattr(klass, "_asdict")
+            and hasattr(klass, "_fields")
+            and name in klass._fields
+        ):
+            return True
+        return False
 
     def _is_show_member(self, name):
         if self.show_inherited_members:
--- a/numpydoc/tests/test_docscrape.py
+++ b/numpydoc/tests/test_docscrape.py
@@ -1694,6 +1694,8 @@ def test_namedtuple_class_docstring():
     assert "MyFoo's class docstring" in sds["Summary"]
     assert len(sds["Attributes"]) == 0
     assert len(sds["Parameters"]) == 2
+    assert sds["Parameters"][0].desc[0] == "The bar attribute"
+    assert sds["Parameters"][1].desc[0] == "The baz attribute"
 
 
 if __name__ == "__main__":
--- a/numpydoc/tests/test_docscrape.py
+++ b/numpydoc/tests/test_docscrape.py
@@ -1641,7 +1641,7 @@ def test_class_docstring_cached_property():
     assert class_docstring["Attributes"][0].name == "val"
 
 
-def test_named_tuple_no_duplicate_attributes():
+def test_namedtuple_no_duplicate_attributes():
     """
     Ensure that attributes of namedtuples are not duplicated in the docstring.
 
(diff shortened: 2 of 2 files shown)


stefanv (approved):

numpydoc/docscrape.py
```diff
@@ -715,6 +716,19 @@ def properties(self):
             )
         ]
 
+    @staticmethod
+    def _should_skip_member(name, klass):
+        if (
+            # Namedtuples should skip everything in their ._fields as the
+            # docstrings for each of the members is: "Alias for field number X"
```
pitrou, line 723: Does it work properly with https://github.com/numpy/numpydoc/issues/257#issuecomment-1002632904 ?

fohrloop: Tested with code from [this comment](https://github.com/numpy/numpydoc/issues/257#issuecomment-1002632904), using

```
.. autoclass:: mypkg.Foo
    :members:
```

that outputs:

image


does that answer to the question?

fohrloop:
As a side note, it might be cleaner to have all "should_skip_member" -type of logic under one method. Right now there is 

```python
                not name.startswith("_") # <----- this
                and not self._should_skip_member(name, self._cls)
                and ( #  <--------- this
                    func is None
                    or isinstance(func, (property, cached_property))
                    or inspect.isdatadescriptor(func)
                )
                and self._is_show_member(name) 

...


    def _is_show_member(self, name):
        if self.show_inherited_members:
            return True  # show all class members
        if name not in self._cls.__dict__: # <------- and perhaps this?
            return False  
        return True

```

which could be moved under one place.

numpydoc/docscrape.py
pitrou: It does, thank you!

rossbar (approved):
Thanks @fohrloop ! I took the liberty of pushing up a test that fails on `main` and passes after this change. This does indeed handle the duplication of the named fields mentioned in #257 .

PS; I was surprised to learn that there is no way to do `isinstance` checks with named tuples - you learn something new everyday!

fohrloop:
Thanks @rossbar for the test! Let me know if there's need for additional tests. For example, could it be tested that namedtuple *with* some custom docs in the params like this works as expected:

```
class SomeClass(NamedTuple):
    """A docstring.
    """
    
    foo: str = 'text'
    """This is a docstring of foo 
    """
```

It was tested manually, but I don't know if it would be beneficial to have such automated test, too. Not sure how to implement that one, though. 

Is there something else that should be done here? Does the code need refactoring..? Or something else?

rossbar:
> For example, could it be tested that namedtuple with some custom docs in the params like this works as expected

That's a good idea, but shouldn't be a blocker for this PR. I'll try to add one today if I have time but don't wait on me - if someone pushes the green button in the meantime that's a positive outcome!

> Is there something else that should be done here? Does the code need refactoring..? Or something else?

No other action necessary IMO - this is good to go. The only reason I didn't merge it myself is because I pushed up some code, so someone should have an opportunity to take a look at that. If it's not in by this evening I'll merge it!

rossbar:
A day late, but I think I managed to design a test that captures your last comment @fohrloop - take a look and LMK what you think. If it looks correct to you I'll go ahead and merge!

fohrloop:
@rossbar awesome, thanks! It seems to test well that the `MyFooWithParams.bar` and `MyFooWithParams.baz` are included in the documentation. Would it make sense to test also that the content of the docstring is there? Something like

```
assert sds["Parameters"][0].desc[0] == "The bar attribute"
```

On the side note I don't fully undestand why the `.bar` and `.baz` are listed under "Parameters" and not "Attributes".

rossbar:
> On the side note I don't fully undestand why the .bar and .baz are listed under "Parameters" and not "Attributes".

This is because they are included under the `Parameters` heading in the class docstring. Without that, they wouldn't be there at all (i.e. `len(sds["Parameters"]) == 0` for the previous test case).

I think all the bases are now covered, so I'll go ahead and get this in, but if there are other corner-cases related to namedtuple/dataclasses, please don't hesitate to continue fleshing them out!

fohrloop:
### myst_parser compatibility

I just did some tests with this using numpydoc 1.7.0rc0.dev0 installed from latest master (commit 46f532a824639a97479039fc122533915cdfa10f):

```
python -m pip install https://github.com/numpy/numpydoc/archive/main.zip
```

It turns out that if you use `numpydoc` with `myst_parser`, `numpydoc` must be *listed before `myst_parser`*  in the `conf.py`;

**This works**:

```python
#conf.py

extensions = [
    "numpydoc",
    "myst_parser",
    ...
]
```

**This will not work**:

```python
#conf.py

extensions = [
    "myst_parser",
    "numpydoc",
    ...
]
```

*(just commenting this here so that if someone has problems with this they know what to look at)*

rossbar:
Thanks for bringing this up @fohrloop - this is worth it's own, separate issue here (and potentially with `myst_parser`) so that any config collisions between extensions can be resolved. Feel free to do so, else I will cp/paste the above to a new issue later.

fohrloop:
Yeah I'm not sure if this is an "issue" or just a feature. Not sure what  `myst_parser` does in between, but I'd guess the document tree (or what it is called) is converted somehow and numpydoc does not understand it anymore? If this is solvable in a way that numpydoc and myst_parser could be placed in either order, that's great, but I can myself live with that ordering. Just hoping that there will be no other issue which would demand the opposite order 😁🤞

And it's likely that this could/should be solved in myst_parser side as it's the one first in the list when there are problems.

PR #532, opened 2024-03-17
PERF: wrap inspect.getsourcelines with cache

dontgoto:
While looking into some docstring checks for pandas I saw the opportunity to shave off about 1s of our remaining 7s runtime of numpydoc functions by wrapping `inspect.getsourcelines` with a cache since it usually gets called twice with the same arguments and it is comparatively expensive.


stefanv (approved):

rossbar (approved):
Seems reasonable to me - it'd be nice to chase down why the full parsing is called twice and whether there's a way to eliminate that, but that's a separate issue. Thanks @dontgoto !

PR #533, opened 2024-03-22
Use trusted publisher

jarrodmillman:
I enabled GH as a trusted publisher for PyPI:
2024-03-22T08:06:39,067261850-07:00

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,34 @@
+name: Build Wheel and Release
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


jarrodmillman:
Once the tests pass, I will self-merge. I am going to release numpydoc `1.7.0rc0`, verify that things work, update the release process notes, test numpydoc `1.7.0rc0` on some projects, and release numpydoc `1.7.0` next week.

Currently, I restrict publishing and we may want to loosen it once I verify that everything works.

Issue #467, opened 2023-06-29
Unify cli validation interfaces


stefanv:
See https://github.com/numpy/numpydoc/pull/454

`python -m numpydoc`
`python -m numpydoc.hooks.validate_docstrings`

These CLI interfaces do slightly different things, but one could imagine a command word interface a-la git.

Also, it would be good to expose this as an executable user-script `numpydoc`.

/cc @stefmolin

stefmolin:
I'll tackle this one after #466

stefmolin:
Is there a preference between these two options?

```shell
$ numpydoc render           # runs the current numpydoc CLI's render_object()
$ numpydoc validate         # runs the current --validate
$ numpydoc validate --ast   # runs the current numpydoc.hooks.validate_docstrings
```

OR

```shell
$ numpydoc                    # runs the current numpydoc CLI's render_object()
$ numpydoc --validate         # runs the current --validate
$ numpydoc --ast-validation   # runs the current numpydoc.hooks.validate_docstrings
```

I'm calling the hook method "ast" because it uses the abstract syntax tree unlike the original validation, but let me know if we should use something else.

stefanv:
How about:

```
$ numpydoc  # alias for numpydoc render
$ numpydoc render
$ numpydoc validate
```

stefmolin:
And `numpydoc validate` would have the flag to trigger the hook version?

stefanv:
Or just drop the alias; it's not like this is done so often by hand.

stefmolin:
Looking at this more, I actually don't think the `--validate` logic should be combined with the hook logic because they operate on different inputs: the hook takes files and the `--validate` option takes a dotted path, which could be to anything.

How about having separate subcommands for each?

```shell
$ numpydoc render
$ numpydoc validate
$ numpydoc hook
```

stefanv:
Nice and clean; I prefer that too.

stefanv:
`hook` is perhaps not a super intuitive name; `numpydoc lint`?

PR #537, opened 2024-03-30
Unify CLIs

stefmolin:
Changes:
- Created `numpydoc` entry point (replacing the hook's entry point).
- Moved CLI logic out of `__main__.py` and into a dedicated `cli.py` file.
- Refactored the CLI to use subparsers and treat each of the unique functionalities as a subcommand like `git`: `render`, `validate`, and `lint`.
- Updated `test_main.py` for changes and addition of `lint` option.
- Removed the `-m` option from the hook, so that usage is now from `numpydoc lint`.
- Updated the entry point for the pre-commit hook to `numpydoc lint`.
- Updated validation docs.
- Updated the `test.yml` workflow, but I didn't add a corresponding `lint` command as I don't fully understand what that part does.

Closes https://github.com/numpy/numpydoc/issues/467

### DIFF
--- a/numpydoc/__main__.py
+++ b/numpydoc/__main__.py
@@ -2,55 +2,6 @@
 Implementing `python -m numpydoc` functionality.
 """
 
-import sys
-import argparse
-import ast
+from .cli import main
 
-from .docscrape_sphinx import get_doc_object
-from .validate import validate, Validator
-
-
-def render_object(import_path, config=None):
-    """Test numpydoc docstring generation for a given object"""
-    # TODO: Move Validator._load_obj to a better place than validate
-    print(get_doc_object(Validator._load_obj(import_path), config=dict(config or [])))
-    return 0
-
-
-def validate_object(import_path):
-    exit_status = 0
-    results = validate(import_path)
-    for err_code, err_desc in results["errors"]:
-        exit_status += 1
-        print(":".join([import_path, err_code, err_desc]))
-    return exit_status
-
-
-if __name__ == "__main__":
-    ap = argparse.ArgumentParser(description=__doc__)
-    ap.add_argument("import_path", help="e.g. numpy.ndarray")
-
-    def _parse_config(s):
-        key, _, value = s.partition("=")
-        value = ast.literal_eval(value)
-        return key, value
-
-    ap.add_argument(
-        "-c",
-        "--config",
-        type=_parse_config,
-        action="append",
-        help="key=val where val will be parsed by literal_eval, "
-        "e.g. -c use_plots=True. Multiple -c can be used.",
-    )
-    ap.add_argument(
-        "--validate", action="store_true", help="validate the object and report errors"
-    )
-    args = ap.parse_args()
-
-    if args.validate:
-        exit_code = validate_object(args.import_path)
-    else:
-        exit_code = render_object(args.import_path, args.config)
-
-    sys.exit(exit_code)
+raise SystemExit(main())
--- a/numpydoc/cli.py
+++ b/numpydoc/cli.py
@@ -1,0 +1,68 @@
+"""The CLI for numpydoc."""
+
+import argparse
+import ast
+from typing import Sequence, Union
+
+from .docscrape_sphinx import get_doc_object
+from .hooks import validate_docstrings
+from .validate import Validator, validate
+
+
+def render_object(import_path: str, config: Union[list[str], None] = None) -> int:
+    """Test numpydoc docstring generation for a given object."""
+    # TODO: Move Validator._load_obj to a better place than validate
+    print(get_doc_object(Validator._load_obj(import_path), config=dict(config or [])))
+    return 0
+
+
+def validate_object(import_path: str) -> int:
+    """Run numpydoc docstring validation for a given object."""
+    exit_status = 0
+    results = validate(import_path)
+    for err_code, err_desc in results["errors"]:
+        exit_status += 1
+        print(":".join([import_path, err_code, err_desc]))
+    return exit_status
+
+
+def main(argv: Union[Sequence[str], None] = None) -> int:
+    """CLI for numpydoc."""
+    ap = argparse.ArgumentParser(prog="numpydoc", description=__doc__)
+    subparsers = ap.add_subparsers(title="subcommands")
+
+    def _parse_config(s):
+        key, _, value = s.partition("=")
+        value = ast.literal_eval(value)
+        return key, value
+
+    render = subparsers.add_parser(
+        "render",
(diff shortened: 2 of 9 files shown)


jarrodmillman (approved):
Thanks!!

jarrodmillman:
Stefan is back and will take a look at this soon. He is a little backlogged. Thanks for your patience.

stefanv:
Thank you very much @stefmolin.

I see that I get a traceback when no import path is specified.

I think I suggested the `numpydoc` being an alias for `numpydoc render` before (bad Stéfan!), and now realize it is rather confusion. Perhaps we should let `numpydoc` (without arguments) print help? Or, at minimum, print a proper error message when `import_path` is None.

numpydoc/__main__.py
```diff
@@ -2,55 +2,6 @@
...
-    )
-    args = ap.parse_args()
-
-    if args.validate:
-        exit_code = validate_object(args.import_path)
-    else:
-        exit_code = render_object(args.import_path, args.config)
-
-    sys.exit(exit_code)
+raise SystemExit(main())
```
stefanv, line 7: Neat, I never knew about SystemExit!

numpydoc/cli.py
```diff
@@ -0,0 +1,68 @@
...
+    subparsers = ap.add_subparsers(title="subcommands")
+
+    def _parse_config(s):
+        key, _, value = s.partition("=")
+        value = ast.literal_eval(value)
+        return key, value
+
+    render = subparsers.add_parser(
+        "render",
+        description="Test numpydoc docstring generation for a given object.",
```
stefanv, line 41: Why does this "Test" docstring generation; doesn't it generate an expanded RST version of the docstring?

numpydoc/cli.py
```diff
@@ -0,0 +1,68 @@
...
+        type=_parse_config,
+        action="append",
+        help="key=val where val will be parsed by literal_eval, "
+        "e.g. -c use_plots=True. Multiple -c can be used.",
+    )
+    render.set_defaults(func=render_object)
+
+    validate = subparsers.add_parser(
+        "validate",
+        description="Validate the docstring with numpydoc.",
```
stefanv, line 57: Can we clarify here the distinction between `validate` and `lint`. They do the same, except the one operates on a function and the other on a file? But `validate` doesn't allow disabling certain rules, which confused me.

numpydoc/cli.py
```diff
@@ -0,0 +1,68 @@
...
+        help="validate the object and report errors",
+    )
+    validate.add_argument("import_path", help="e.g. numpy.ndarray")
+    validate.set_defaults(func=validate_object)
+
+    lint_parser = validate_docstrings.get_parser(parent=subparsers)
```
stefanv, line 63: How is the lint_parser hooked up? Why not the same way as render and validate?

Also, the lint docstring on my machine contains:

```
Check codes to ignore. Currently ignoring the following from /home/stefan/src/numpy/pyproject.toml: Values provided here will be
                        in addition to the above, unless an alternate config is provided.
```

Not quite right?

numpydoc/hooks/validate_docstrings.py
```diff
@@ -357,10 +370,21 @@ def main(argv: Union[Sequence[str], None] = None) -> int:
         + "\n"
     )
 
-    parser = argparse.ArgumentParser(
-        description="Run numpydoc validation on files with option to ignore individual checks.",
-        formatter_class=argparse.RawTextHelpFormatter,
+    description = (
+        "Run numpydoc validation on files with option to ignore individual checks."
     )
+    if parent is None:
```
stefanv, line 376: I guess this is the part that confused me; why have two ways of adding the subparser? And, worse, without a parent automatically add lint subparser?

stefanv (commented):
Looks good, I think the `lint` subparser perhaps needs to be hooked up more cleanly, but tested everything locally and it works. Thank you again!

numpydoc/cli.py
stefmolin: This actually comes from the original docstring for `render_object()`, which was the only thing called for the original CLI's behavior:

https://github.com/numpy/numpydoc/blob/3a21e5440c90a0f306c42fdc303083a12927969b/numpydoc/__main__.py#L14

numpydoc/cli.py
stefmolin: I think it works for non-functions as well. How about this?

> Validate an object's docstring conforms to the numpydoc standard.

numpydoc/cli.py
stefanv: :shrug: That's weird... but there are lots of weird things in this code :) Feel free to change whatever you see fit, or if it feels out of scope that's fine too.

numpydoc/cli.py
stefanv: "Validate that an object's docstring conforms to the numpydoc standard" :+1:

stefanv: Or, "Determine (or check?) whether an object's docstring..."

stefanv: Otherwise just: "Validate an object's docstring against the numpydoc standard"

stefanv: I may prefer the last version, but anything gramatically sound is fine.

stefanv: > I'm not sure what to do about the second part of your comment. If you run `numpydoc validate --help` you see what can be passed as arguments, and there aren't any for disabling rules.

This was more of a question: *should* there be the option of disabling certain rules? Why are these not exactly equivalent in their functionality?

stefmolin: > > I'm not sure what to do about the second part of your comment. If you run `numpydoc validate --help` you see what can be passed as arguments, and there aren't any for disabling rules.
> 
> This was more of a question: _should_ there be the option of disabling certain rules? Why are these not exactly equivalent in their functionality?

While I agree this should eventually be the case, this definitely seems out of scope for this PR, but at least the separate subcommands will not draw attention to arguments that have no affect like before (see https://github.com/numpy/numpydoc/issues/459).

numpydoc/cli.py
stefmolin: There was a bug with this hack. I will move things around.

numpydoc/cli.py
stefmolin: I will use a version of what you put above. This PR has lots of potential for snowballing, and I want to keep it as focused as possible.

stefanv (approved):
Thanks @stefmolin! The new interface is great.

stefanv:
@rossbar @jarrodmillman handing it over to you.

PR #538, opened 2024-03-30
Fix typo in label-check.yml

stefmolin:
I thought maybe that was why the check fails, but it seems to be something wrong with line 20, which shows up as incorrect syntax in my editor.


stefanv:
Well, that's a mistake for sure.

PR #547, opened 2024-04-27
Test on sphinx 7.3

jarrodmillman:
Undoing pins from #542 now that sphinx 7.3.7 is out which should fix the issue.


rossbar (approved):
CI is green, so 7.3.7 seems to have done the trick!

PR #548, opened 2024-04-27
Classify development status as Production/Stable

jarrodmillman:


rossbar (approved):
Nice catch - happy to put this in now or wait if there are other tweaks to the project metadata on deck, just LMK!

PR #549, opened 2024-04-27
Clean up old sphinx cruft

jarrodmillman:
I don't think we really need either of these.


rossbar (approved):
Agreed, thanks @jarrodmillman !

PR #555, opened 2024-05-11
Use ruff for linting and formatting

jarrodmillman:
A couple of issues:

1. We don't appear to be testing `BadExamples` in `numpydoc/tests/test_validate.py`.
2. I had to fix the line numbers (see 2e6d88b) to get the prerelease tests to pass, but it wasn't necessary for the standard tests. For the standard tests we run `pytest -v --pyargs numpydoc`. For the prerelease test we run `pytest -v --pyargs .`.

`numpydoc/tests/hooks` is in the source directory:
```
$ ls -1 numpydoc/tests/
hooks
test_docscrape.py
test_full.py
test_main.py
test_numpydoc.py
test_validate.py
test_xref.py
tinybuild
```
but isn't in the installed package
```
$ ls -1 ~/.venv/nx/lib64/python3.11/site-packages/numpydoc/tests/
test_docscrape.py
test_full.py
test_main.py
test_numpydoc.py
test_validate.py
test_xref.py
tinybuild
```

I ignore a large number of linting checks because they don't have automated fixes (or their automated fixes are unsafe). I think we should eventually enable them one by one and implement the fixes in follow-up PRs. I would like to keep them separate because they will require more attention and careful review than these changes. I would like to get this PR in relatively quickly because I think moving to ruff is worth it. It is much faster, implements multiple tools,  is being rapidly developed, and has been adding features specific for our community.


numpydoc/tests/test_numpydoc.py
```diff
@@ -142,15 +142,14 @@ def test_clean_text_signature():
...
-@pytest.fixture
+@pytest.fixture()
 def f():
     def _function_without_seealso_and_examples():
         """
         A function whose docstring has no examples or see also section.
 
         Expect SA01 and EX01 errors if validation enabled.
         """
-        pass
```
stefanv, line 153: I did not know a function with only a docstring is valid!

stefanv (approved):
Looks good to me; I agree that it's better to get the toolchain in place quickly, and to the iteratively refine.

PR #558, opened 2024-05-22
See Also Parser Recognizes Sphinx XREF

j9ac9k:
The See Also section raises a ValueError when it encounters a valid sphinx cross-reference such as 

```rst
:meth:`QImage.save <QImage.save>
```

This commit allows the regex to parse the sphinx target, which is the section between the angled brackets.

Author Note: This is my first contribution to a repo such as this, so I likely missed something, comments/nit-picks welcome.


stefanv:
We should probably allow this, but it would need a test to prevent future regression.

j9ac9k:
> We should probably allow this, but it would need a test to prevent future regression.

Happy to add tests, I didn't notice if there was an existing test I should add or, or just roll my own thing, which I can do too, but would likely need some guidance on where it should go.

PR #564, opened 2024-06-03
Make regenerating requirements unnecessary

Carreau:
Point back at pyproject.toml this might avoid a pre-commit and avoid out of sync things.


Carreau:
Next step would be to remove the actual reference to those files to `.[<target>]` then remove those files.

Carreau:
See https://github.com/scikit-image/scikit-image/pull/7435#discussion_r1626669399

this is not strictly equivalent as it also install the current version of numpydoc.

PR #565, opened 2024-06-11
DOC: Extend explanation on single vs double backticks.

Carreau:
Add the special case for constant, and try to separate semantic (can link), to style (monospace, link)

See also #525

### DIFF
--- a/doc/format.rst
+++ b/doc/format.rst
@@ -753,9 +753,8 @@ monospaced font; in reST, *double* backticks are for ``monospaced font``,
 whereas the behavior of single backticks is defined by the default role. This
 leads to the following style recommendations:
 
-- Any object that can be linked to should be enclosed in single backticks. That
-  is to say : Module, class, function, method, and attribute names are some of
-  the objects that should be enclosed in single backticks. On a case by cases,
+- For example, module, class, function, method, and attribute names
+  should be enclosed in single backticks. On a case-by-case basis,
   instances may also be included in single backticks, like :py:obj:`None`,
   :py:obj:`True`, :py:obj:`False`.
 
--- a/doc/format.rst
+++ b/doc/format.rst
@@ -753,18 +753,26 @@ monospaced font; in reST, *double* backticks are for ``monospaced font``,
 whereas the behavior of single backticks is defined by the default role. This
 leads to the following style recommendations:
 
-- Module, class, function, method, and attribute names should render as
-  hyperlinks in monospaced font (e.g. :any:`numpy`); depending on project
-  settings, this may be accomplished simply be enclosing them in single
-  backticks. If the hyperlink does not render as intended, explicitly
-  include the appropriate role and/or namespace.
+- Any object that can be linked to should be enclosed in single backticks. That
+  is to say : Module, class, function, method, and attribute names are some of
+  the objects that should be enclosed in single backticks. On a case by cases,
+  instances may also be included in single backticks, like :py:obj:`None`,
+  :py:obj:`True`, :py:obj:`False`.
+
+  These will typically render as hyperlinks in monospaced font (e.g.
+  :any:`numpy`) to the relevant object documentation; depending on project
+  settings. If the hyperlink does not render properly or point to the intended
+  object, explicitly include the appropriate role and/or namespace.
+
 - This guide continues to recommended that parameter names be enclosed within
   single backticks. Currently, this may cause parameter names to render
   improperly and cause warnings, but numpydoc will soon release a feature
   that causes them to render as monospaced hyperlinks to the parameter
   documentation.
-- All other text that is intended to render in ``monospaced`` font should be
-  enclosed within ````double backticks````.
+
+- All other text that is intended to render in ``monospaced`` font and is not
+  supposed to refer to another docs page should be enclosed within ````double
+  backticks````.
 
 A more extensive example of reST markup can be found in `this example
 document <http://docutils.sourceforge.net/docs/user/rst/demo.txt>`_;


doc/format.rst
```diff
@@ -753,18 +753,26 @@ monospaced font; in reST, *double* backticks are for ``monospaced font``,
...
 leads to the following style recommendations:
 
-- Module, class, function, method, and attribute names should render as
-  hyperlinks in monospaced font (e.g. :any:`numpy`); depending on project
-  settings, this may be accomplished simply be enclosing them in single
-  backticks. If the hyperlink does not render as intended, explicitly
-  include the appropriate role and/or namespace.
+- Any object that can be linked to should be enclosed in single backticks. That
+  is to say : Module, class, function, method, and attribute names are some of
+  the objects that should be enclosed in single backticks. On a case by cases,
```
mdhaber, line 758: I think "for example" is more appropriate than "that is" here because it gives examples rather than restating. Also, I assume you meant "on a case-by-case basis"
```suggestion
- For example, module, class, function, method, and attribute names
  should be enclosed in single backticks. On a case-by-case basis,
```

Maybe we do legitimately disagree a little bit. I do think that text that refers to code should always be monospaced. For objects that happen to link, the way to accomplish this is to enclose them in single backticks.

doc/format.rst
```diff
@@ -753,18 +753,26 @@ monospaced font; in reST, *double* backticks are for ``monospaced font``,
...
 - This guide continues to recommended that parameter names be enclosed within
   single backticks. Currently, this may cause parameter names to render
   improperly and cause warnings, but numpydoc will soon release a feature
   that causes them to render as monospaced hyperlinks to the parameter
   documentation.
-- All other text that is intended to render in ``monospaced`` font should be
-  enclosed within ````double backticks````.
+
+- All other text that is intended to render in ``monospaced`` font and is not
+  supposed to refer to another docs page should be enclosed within ````double
```
mdhaber, line 774: I thought "and is not supposed to refer to another docs page" was implied because this says "All *other* text" and it appears after we've discussed links. Does it need to be more explicit?

mdhaber:
I think the original sentence and this one focus on separate things.
- Things that link should be enclosed in single backticks.
- Things that are code should be rendered in monospaced font.

This exchanges one for the other. Maybe the best would be to more clearly distinguish between these two ideas and mention that often things that link are code, and links to objects appear in monospaced font, so in these cases single backticks take priority because they accomplish both.

doc/format.rst
Carreau: I think we can also be more explicit, and you can have monospaced linked to another page if you use a directive. like `` :ref:`the text <the_ref>` `` technically I think. I'm just also trying to separate rendering from semantic.

Carreau: But I'm happy to remove.

Carreau:
> I think the original sentence and this one focus on conceptualy disting things.
> 
> * Things that link should be enclosed in single backticks.
> * Things that are code should be rendered in monospaced font.
> 
> This exchanges one for the other. Maybe the best would be to more clearly distinguish between these two ideas and mention that often things that link are code, and links to objects appear in monospaced font, so in these cases single backticks take priority because they accomplish both.

But isn't that conflating again semantic and styling ? Say the community realize that double backticks is much better in blinking red, and link in rainbow underline italic ?

Can we find a way to drop any reference to the _visual_ representation ? can we replace `monospace` by code sample ? The fact that it is monospace is again purely a css-theme thing (for html), and if you render in terminal well... everything is monospace.

mdhaber:
> Can we find a way to drop any reference to the visual representation

I think our motivations are different. I am actually interested in the visual representation. If there were some other visual cue that were to designate text as "code sample" and it were in proportional font, I would probably comment that it should be in monospaced font.

I suppose if all IDEs and terminals began to use proportional font and chose some other way to designate code as code, my brain might change, too. But while unicode seems to be (is?) the universal convention for "this is code", I don't hesitate to refer to the appearance.

stefanv:
We have a translation layer, in the form of numpydoc, so that we can define a format of our choice and render that however we want. So, I also lean towards focusing on the semantics, and then implementing whichever rendering we want through the library.

E.g., while we are currently tightly coupled to sphinx, one can easily imagine a nearby future in which we render docs with papyri, or a combo of papyri and mystmd.

Carreau:
I agree personally with @mdhaber wish for final style, but as numpydoc does not control the rendering, only the parsing, I think that using the style as the main reason is a weaker argument than semantic, because if a project disagrees on monospace, why would they follow single backticks and not use stars ? While if you recommend a semantic and note that this semantic is usually monospace, to do not give a reason for a project to deviate from but simply to change the theme. 

Maybe we should get clearer distinction about recommended syntax and semantic. recommended formatting and style.

Carreau:
Another argument against using the final monospace style, is that currently we all wish for both single backticks and double backticks to render as monospace font, the only difference being that single backtick will also be link to the relevant objects (or have a hover tooltip maybe ?) who knows.

So I think I find it weird because my brain reads the current proposal as "single backticks should be used for targets that will render as monospace, while you should use double backticks for code that should render monospace".

doc/format.rst
mdhaber: Oops, I think I messed up here. I didn't mean to remove the first part: "Any object that can be linked to should be enclosed in single backticks." Otherwise, the change looks like what I had in mind.

doc/format.rst
mdhaber: I read it as redundant, but if others don't, it's fine!

mdhaber:
I guess this still says that things "typically" appear in monospace. Others have a better idea of the scope of numpydoc, so this is fine with me. Two last comments: 

- It says "On a case-by-case basis, instances may also be included in single backticks, like :py:obj:`None`, :py:obj:`True`, :py:obj:`False`." Is "instances" synonymous with "literals"? Is it recommended for them to be in double backticks otherwise?
- We mentioned "Module, class, function, method, and attribute names" but I forgot about _returns_. Should those get single backticks, too? They are sort of an interesting case because we give these labels to the returned values, but those labels don't necessarily appear in any code.

mdhaber:
Oops, fell off the radar. Do you still want this in? I can merge as is.

Carreau:
Reopening in case you want to merge, I don't have any particular agenda, just trying to clean up a bit my list of task and reduce mental load.

mdhaber:
Oops! My mistake - I don't have commit rights in the numpydoc repo! I guess that's why I didn't merge in the first place : )

stefanv:
> Oops! My mistake - I don't have commit rights in the numpydoc repo!

That should be fixed now.

doc/format.rst
```diff
@@ -755,18 +755,25 @@ monospaced font; in reST, *double* backticks are for ``monospaced font``,
...
 
-- Module, class, function, method, and attribute names should render as
-  hyperlinks in monospaced font (e.g. :any:`numpy`); depending on project
-  settings, this may be accomplished simply be enclosing them in single
-  backticks. If the hyperlink does not render as intended, explicitly
-  include the appropriate role and/or namespace.
+- For example, module, class, function, method, and attribute names
+  should be enclosed in single backticks. On a case-by-case basis,
+  instances may also be included in single backticks, like :py:obj:`None`,
+  :py:obj:`True`, :py:obj:`False`.
```
mdhaber, line 761: @Carreau this renders like:
image

Seeing only the rendered version and not the underlying reST, I would assume from the text that these object names only need to be included in single backticks to render like this, but that's not quite right. Should these go at the end of the paragraph, after the idea of explicitly including the role and/or namespace is mentioned?

doc/format.rst
```diff
@@ -755,18 +755,25 @@ monospaced font; in reST, *double* backticks are for ``monospaced font``,
...
+
+  These will typically render as hyperlinks in monospaced font (e.g.
+  :any:`numpy`) to the relevant object documentation; depending on project
+  settings. If the hyperlink does not render properly or point to the intended
```
mdhaber, line 765:
```suggestion
  :any:`numpy`) to the relevant object documentation.
  If the hyperlink does not render properly or point to the intended
```
Semicolon is no longer the appropriate punctuation, and I think this part can just be removed now.

mdhaber (commented):
Thanks @stefanv for the commit right!

@Carreau just one question for you

PR #566, opened 2024-06-12
Do not rely on requirements.txt in ci, use .[test,doc]

Carreau:
This is one step toward not needing those file and having one less command to sugest to users.


Carreau:
Alternative/complementary to #564

larsoner (approved):
Looks like good incremental progress!

Carreau:
thanks !

PR #569, opened 2024-06-20
Fix navbar for documentation pages

melissawm:
I noticed something went wrong and two different search fields were showing in the top navbar for the docs website, causing the navbar items to wrap. This PR removes the extra search field.

Before:

Captura de imagem_20240620_160713

This PR:

Captura de imagem_20240620_160743


rossbar (approved):
Thanks @melissawm !

PR #571, opened 2024-07-08
Move "Attributes" and "Methods" below "Parameters"

lucascolley:
Reorder the sections in `docscrape` to match the style guide. As per https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring, the Attributes section should be "located below the Parameters section", which I am interpreting as directly below. It seems wrong for Attributes to be below Notes, References and Examples.

x-ref https://github.com/scipy/scipy/issues/21099

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -120,17 +120,17 @@ class NumpyDocString(Mapping):
         "Summary": [""],
         "Extended Summary": [],
         "Parameters": [],
+        "Attributes": [],
+        "Methods": [],
         "Returns": [],
         "Yields": [],
         "Receives": [],
+        "Other Parameters": [],
         "Raises": [],
         "Warns": [],
-        "Other Parameters": [],
-        "Attributes": [],
-        "Methods": [],
+        "Warnings": [],
         "See Also": [],
         "Notes": [],
-        "Warnings": [],
         "References": "",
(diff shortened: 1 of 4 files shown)


lucascolley:
this test https://github.com/numpy/numpydoc/blob/54dbc890fdecc4c2e170e208ed2edc05bb62202a/numpydoc/tests/test_docscrape.py#L1258 actually expects Examples to be above Attributes and Methods. So maybe the question of where they should go isn't settled. I think that we should either change the order in the code/tests (my preference - I don't see why See Also should be above Attributes, for example), or change the wording in the style guide.

lucascolley:
I opened gh-572 in case we need higher-level discussion.

rossbar (commented):
FWIW I'm in favor of the change - class attributes and methods are typically key pieces of API, so ensuring that they render near the top of the docstring instead of being buried under the other sections seems like an improvement.

lucascolley:
could someone approve CI please? :)

larsoner:
Done!

lucascolley:
apologies for the iteration, I was being lazy and not running the tests - they all pass locally for me now!

larsoner (approved):
Seems like everyone is in favor so in it goes, thanks @lucascolley !

lucascolley:
thanks all!

Issue #561, opened 2024-05-24
See Also missing reported for module level docstring


stefanv:
`numpydoc lint foo.py` with `foo.py` containing:

```
"This is a file."

```

Same holds for Extended Summary and example section. I.e., ES01, SA01, EX01.

PR #593, opened 2024-12-14
ignore some errors at module level

lucascolley:
closes gh-561

cc @stefanv

### DIFF
--- a/numpydoc/__main__.py
+++ b/numpydoc/__main__.py
@@ -1,5 +1,5 @@
 """
-Implementing `python -m numpydoc` functionality.
+Implementing `python -m numpydoc` functionality
 """
 
 from .cli import main
--- a/numpydoc/__main__.py
+++ b/numpydoc/__main__.py
@@ -1,6 +1,6 @@
 """
 Implementing `python -m numpydoc` functionality
-"""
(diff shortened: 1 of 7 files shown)


numpydoc/__main__.py
```diff
@@ -1,5 +1,5 @@
 """
-Implementing `python -m numpydoc` functionality.
+Implementing `python -m numpydoc` functionality
```
larsoner, line 2: I think all docstrings (including module-level?) typically end with a period, no? In that sense this is a tiny regression. If it's for testing purposes (looks like it?) then a little comment here saying that's the case would help the next person not (mistakenly) fix it.

larsoner (approved):
Just one tiny comment from me.

@stefanv did you want to look?

numpydoc/__main__.py
lucascolley: Yeah, I can add a comment. Also happy to change to a different "defect" if we want.

lucascolley:
thanks @larsoner, addressed!

stefanv (approved):

stefanv:
Thanks, Lucas!

PR #596, opened 2024-12-17
don't pass maxsplit as positional arg

drammock:
Py3.13 introduces a warning when passing `re.split(..., maxsplit)` as positional. This PR changes it to a kwarg.

xref: https://github.com/python/cpython/pull/107778


stefanv (approved):

stefanv:
Thanks, Dan!

PR #597, opened 2024-12-19
Fix *args as positional arguments

mwtoews:
This PR fixes the wording to describe `*args` as _positional_ arguments, and adds a full stop at the end.

### DIFF
--- a/doc/format.rst
+++ b/doc/format.rst
@@ -228,7 +228,7 @@ When documenting variable length positional, or keyword arguments, leave the
 leading star(s) in front of the name and do not specify a type::
 
   *args
-      Additional arguments should be passed as keyword arguments
+      Additional arguments should be passed as positional arguments.
   **kwargs
       Extra arguments to `metric`: refer to each metric documentation for a
       list of all possible arguments.


doc/format.rst
```diff
@@ -228,7 +228,7 @@ When documenting variable length positional, or keyword arguments, leave the
 leading star(s) in front of the name and do not specify a type::
 
   *args
-      Additional arguments should be passed as keyword arguments
+      Additional arguments should be passed as positional arguments.
   **kwargs
       Extra arguments to `metric`: refer to each metric documentation for a
       list of all possible arguments.
```
stefanv, line 234: Thanks for highlighting this confusing phrasing, @mwtoews.

Since this example is entirely made up, I was wondering if we can't come up with something more intuitive. E.g.:

```
def red_print(*args, **kwargs):
   """Print text in red.

   Parameters
   ----------
   *args
       Positional arguments, passed to `print`.
   **kwargs
       Keyword arguments, passed to `print`.
   ...
```

This is not the best example, but it's late :) Something more computational would be good.

I think this was the original goal with the `metric` usage in `**kwargs`, but it doesn't really come across.

PR #601, opened 2025-01-15
ENH: Add support for type annotations

### LINKED ISSUE #196: Support type annotations (PEP 484)
With support for Python 2.7 being dropped in a number of projects, some might consider using type annotations from [PEP 484](https://www.python.org/dev/peps/pep-0484/).

In particular, see discussions in https://github.com/numpy/numpy-stubs, https://github.com/pandas-dev/pandas/issues/14468 https://github.com/scipy/scipy/issues/9038 https://github.com/scikit-learn/scikit-learn/issues/11170 

A recurrent issue when using type annotations is that one has then to keep consistency between types in the docstring and type annotations which is extra work and can be tedious. 

It would be great if numpydoc could take the type information from annotations. For instance,
```py
def func(x):
    """Some function

    Parameters
    ----------
    x : int
        Description of parameter `x`.
    """
```

could optionally be written as,
```py
def func(x: int):
    """Some function

    Parameters
    ----------
    x
        Description of parameter `x`.
    """
```
and generate the same HTML.

To keep backward compatibility this mechanism could be enabled with an optional config flag, and only be used if the type is not provided in the docstring.

For docstring formatting without numpydoc this functionality is implemented in [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/) (that also works with sphinx.napoleon). So maybe integration with that extension could be a possibility.

This would also have the added benefit that type annotations, unlike docstrings types, can be programmatically checked for consistency (e.g. with mypy)..

RastislavTuranyi:
Resolves #196 

Hi! I ran into the issue that numpydoc does not support typehints before Christmas, so I decided to have a crack at it. I have tried multiple different approaches (not in this git history), especially trying to combine numpydoc with [sphinx-autodoc-typehints](https://github.com/tox-dev/sphinx-autodoc-typehints/tree/main), but what I finally settled on is to make use of numpydoc's powerful class system (I have found that the data representation of `numpydoc` and `sphinx-autodoc-typehints` are largely incompatible).

My implementation edits the `_parse_param_list` method - responsible for creating the docstrings for `Parameters`, `Other Parameters`, `Attributes`, `Methods`, `Returns`, `Yields`, `Raises`, and `Receives` sections - so that a type can be obtained from the type hint if one is not specified in the docstring (btw, I believe it should be relatively simple to implement #356 following a similar method). This means that the type hints can be supported for all those sections with minimal code. The way this works is defined in a new method, `_get_type_from_signature`, which by default on `NumpyDocString` only returns an empty string (no type). The actual type inference is implemented on the subclasses:

**`FunctionDoc`**

Getting type hints for functions is the easier task since we can rely fully on `inspect.signature`. Firstly, the signature is obtained in `__init__` to avoid recomputing it at every iteration of `_parse_param_list`. All that is then necessary in `_get_type_from_signature` is to return the type hint associated with the argument name, provided that the function has a signature and that the type hint is set.

**`ClassDoc`**

At the basic level, the type hints for classes are similar to `FunctionDoc` and the code is therefore repeated (there is potentially room for optimisation where the code could be pulled out to `NumpyDocString`, should that be desirable). However, there is an additional complication in that the class documentation may contain the attributes (which may be `@property`) and therefore are not part of the `__init__` signature. This required an additional method, `_find_type_hints`:

- First is handled the case where the `arg_name` is an attribute with a type annotation (`typing.get_type_hints` handles unwrapping etc.)
    - The second `try` `except` block handles the case where the annotation is a normal class (by getting `__name__`) or an annotation from `types`.

```
        type_hints = get_type_hints(cls)
        try:
            annotation = type_hints[arg_name]
        except KeyError:
            ...

        try:
            return str(annotation.__name__)
        except AttributeError:
            return str(annotation)
```

- In the `...`, the remaining options are that the `arg_name` doesn't exist, handled via:

```
            try:
                attr = getattr(cls, arg_name)
            except AttributeError:
                return ""
```

- Or that it is a callable, in which case the return annotation is found via `inspect.signature`:

```
            attr = attr.fget if isinstance(attr, property) else attr

            if callable(attr):
                try:
                    signature = inspect.signature(attr)
                except ValueError:
                    return ""

                if signature.return_annotation == signature.empty:
                    return ""
                else:
                    return str(signature.return_annotation)
```

- Or that it is a class variable without an annotation (but with the value set), in which case it can be extracted:

```
            else:
                return type(attr).__name__
```

*Robustness*

From my testing, this implementation is able to handle everything my other projects could throw at it, but I would not be surprised if there are edge cases where it breaks down. I have not yet written any unit tests since I wasn't sure what the best way to do that in this repository is, or whether this is the method for solving #196 that you want to implement. Regardless, though, I think it might be possible to increase the robustness of my implementation by using some of the type-extraction functions from `sphinx-autodoc-typehints` (whether by depending on it as a whole, or by copying the relevant functions, I am unsure), such as [`get_annotation_args`](https://github.com/tox-dev/sphinx-autodoc-typehints/blob/3af8fb6f6a681d00e88c422bdc5e18ed0158c0b4/src/sphinx_autodoc_typehints/__init__.py#L124) or [`format_annotation`](https://github.com/tox-dev/sphinx-autodoc-typehints/blob/3af8fb6f6a681d00e88c422bdc5e18ed0158c0b4/src/sphinx_autodoc_typehints/__init__.py#L183).

**Conclusion**

Please let me know what you think! I am keen to get `numpydoc` working with type annotations, and I believe that this implementation is a straightforward way to do that, but I am curious what your plans are for this issue.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -9,6 +9,7 @@ import textwrap
 from collections import namedtuple
 from collections.abc import Callable, Mapping
 from functools import cached_property
+from typing import ForwardRef, get_type_hints
 from warnings import warn
 
 
@@ -666,7 +667,12 @@ class ClassDoc(NumpyDocString):
             doc = pydoc.getdoc(cls)
 
         if cls is not None:
-            self._signature = inspect.signature(cls.__init__)
+            try:
+                self._signature = inspect.signature(cls.__init__)
+            except ValueError:
+                self._signature = None
+        else:
+            self._signature = None
 
         NumpyDocString.__init__(self, doc)
 
@@ -752,14 +758,50 @@ class ClassDoc(NumpyDocString):
         )
 
     def _get_type_from_signature(self, arg_name: str) -> str:
-        try:
-            parameter = self._signature.parameters[arg_name.replace("*", "")]
-            if parameter.annotation == parameter.empty:
-                return ""
-            else:
-                return str(parameter.annotation)
-        except AttributeError:
+        if self._signature is None:
             return ""
+
+        arg_name = arg_name.replace("*", "")
+        try:
+            parameter = self._signature.parameters[arg_name]
+        except KeyError:
+            return self._find_type_hint(self._cls, arg_name)
+
+        if parameter.annotation == parameter.empty:
+            return ""
+        else:
+            return str(parameter.annotation)
+
+    @staticmethod
+    def _find_type_hint(cls: type, arg_name: str) -> str:
+        type_hints = get_type_hints(cls)
+        try:
+            annotation = type_hints[arg_name]
+        except KeyError:
+            try:
+                attr = getattr(cls, arg_name)
+            except AttributeError:
+                return ""
+
+            attr = attr.fget if isinstance(attr, property) else attr
+
+            if callable(attr):
+                try:
+                    signature = inspect.signature(attr)
+                except ValueError:
+                    return ""
+
+                if signature.return_annotation == signature.empty:
+                    return ""
+                else:
+                    return str(signature.return_annotation)
+            else:
+                return type(attr).__name__
+
+        try:
+            return str(annotation.__name__)
+        except AttributeError:
+            return str(annotation)
 
 
 def get_doc_object(
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -581,7 +581,11 @@ def dedent_lines(lines):
 class FunctionDoc(NumpyDocString):
     def __init__(self, func, role="func", doc=None, config=None):
         self._f = func
-        self._signature = inspect.signature(func)
+        try:
+            self._signature = inspect.signature(func)
+        except ValueError:
+            self._signature = None
+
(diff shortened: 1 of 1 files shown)


RastislavTuranyi:
Assuming no news is good news, I have gone ahead and wrote the missing tests (as well as fixed the failing ones)! There were also some changes in response to test failures (original PR updated), the biggest of which concerns method types and class attribute type inference:

- Previously, the type hint inserted for:
  - methods in the methods section would be the return type of the method, and
  - class attributes without type hint would be:
    - the return type if the value is a callable (function, method, class, etc.)
    - the type of the value otherwise
- Now, the type hint inserted for:
  - methods in the methods section is `function`, and
  - class attributes without type hint is the type of the value

Now, this should be an improvement since as far as I can see, the type hints for methods are not used at all, so it doesn't seem to matter what they are and the type inference is just better, but it raises two important questions:

- Should we be adding a type to the methods at all?
- Should we be doing the type inference if class attributes have not been type-hinted?

Issue #591, opened 2024-12-09
Check GL08 on class __init__ constructors


mattgebert:
I was looking for a way to skip GL08 checks on class constructors (`__init__`) throughout my code. This is particularly as the numpydoc recommendation for class strings documents the constructor:

> **Class docstring**
> Use the same sections as outlined above (all except [Returns](https://numpydoc.readthedocs.io/en/latest/format.html#returns) are applicable). The constructor (\_\_init\_\_) should also be documented here, the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section of the docstring details the constructor’s parameters.

It seems that there ought to be a default check ignore for `__init__` constructors if no docstring has been provided (and if a class docstring has been provided). Obviously if an (optional) docstring for `__init__` has been created, then it ought to be checked.

This conditional checking doesn't seem possible with the default `override_<code>`, `exclude`  toml tags.
Keen to hear others thoughts!

mattgebert:
In case it wasn't clear, `GL08` has the description "The object does not have a docstring".

mattgebert:
Hey all (@larsoner, @stefanv), 

I have re-opened the issue as the current implementation is not working with the `AstValidator`. It is my understanding that the `AstValidator` is the preferred validator class, used when running numpydoc linting and CI testing? (i.e. The modifications we made are not /will not be used in the majority of numpydoc validation).

When we improved the class constructor __init__ GL08 reporting, we only wrote tests for the `Validator` class which works perfectly. This is because  `Validator` uses the `inspect` library to find the code_obj, we can determine the parent module / class to check if the `__init__` is documented in the constructor.

However, under the implementation of the abstract syntax tree (ast) validator (`numpydoc.hooks.validate_docstrings.AstValidator`) the `code_obj` attribute does not become defined, making the resolving of the parent class (that the constructor is defined on) difficult. I have done a preliminary check if I can still resolve the similar names required to run the `Validator._load_obj` without the `code_obj` attribute, but I haven't figured out a way just yet.

Here's the way we get currently (roughly) get the class reference from the constructor:

    if hasattr(doc, "code_obj"):
        cls_name = doc.code_obj.__qualname__.split(".")[0]
        module_name = doc.code_obj.__module__
        cls = Validator._load_obj(f"{module_name}.{cls_name}")

[...]

PR #605, opened 2025-02-15
Fix typo in validation.rst

stefmolin:
The docs now reference a non-existent rule (G08), which, based on https://github.com/numpy/numpydoc/pull/592, I believe should be GL08.

I also think that line is a little misleading since that does not apply to the pre-commit hook or `numpydoc lint`, which both use the `AstValidator`.

### DIFF
--- a/doc/validation.rst
+++ b/doc/validation.rst
@@ -184,7 +184,7 @@ inline comments:
             pass
 
 Note that a properly formatted :ref:`class <classdoc>` docstring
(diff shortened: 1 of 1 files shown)


stefanv (approved):

mattgebert:
I have just recently realized the hook runs on the AST, and will providing a new PR shortly (re-opened  #591). Just figuring out how to write a test similar to the ones in `test_validate_hook.py`.

PR #607, opened 2025-02-19
Require Sphinx 7

jarrodmillman:


stefmolin:
We should also update the README. It currently says "numpydoc requires Python 3.10+ and sphinx 6+."

PR #613, opened 2025-03-11
BUG: Correct functionality of numpydoc SS05

maxinelasp:
Currently SS05 checks for infinitive verbs by checking if the last character is "s". However, if the first word is "process", which is valid, this incorrectly throws a numpydoc validation error.

This commit allows the first word to end in two "s"es without flagging as an error. This will allow for words like "process" or "progress," while still checking for invalid words like "creates."

I wasn't sure if I should open an issue for this too, I thought it would be easiest to just find the error and submit a PR, but I'm happy to open an issue as well.

### DIFF
--- a/numpydoc/validate.py
+++ b/numpydoc/validate.py
@@ -711,7 +711,12 @@ def validate(obj_name, validator_cls=None, **validator_kwargs):
             errs.append(error("SS03"))
         if doc.summary != doc.summary.lstrip():
             errs.append(error("SS04"))
-        elif doc.is_function_or_method and doc.summary.split(" ")[0][-1] == "s":
+        # Heuristic to check for infinitive verbs - shouldn't end in "s"
+        elif (
+            doc.is_function_or_method
+            and doc.summary.split(" ")[0][-1] == "s"
+            and doc.summary.split(" ")[0][-2] != "s"
+        ):
             errs.append(error("SS05"))
         if doc.num_summary_lines > 1:
             errs.append(error("SS06"))
--- a/numpydoc/validate.py
+++ b/numpydoc/validate.py
@@ -714,6 +714,7 @@ def validate(obj_name, validator_cls=None, **validator_kwargs):
         # Heuristic to check for infinitive verbs - shouldn't end in "s"
         elif (
             doc.is_function_or_method
+            and len(doc.summary.split(" ")[0]) > 1
             and doc.summary.split(" ")[0][-1] == "s"
             and doc.summary.split(" ")[0][-2] != "s"
         ):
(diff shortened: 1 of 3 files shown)


stefanv:
Hah, so iffy what we're doing here, so I think your suggestion is fine. Given how non-robust this test is, we should probably add a comment explaining our lightweight checker mechanism.

pre-commit is complaining; happy to run that for you if you have issues running it locally (`pi p install pre-commit; pre-commit install` once, and you should be good to go.)

maxinelasp:
Apologies, I didn't see that there was a pre-commit in the project! Looks like it's passing now.

stefanv:
Great, thanks, can you add a comment to explain what that clause does? Can be as simple as "# Heuristic to check for infinitive verbs". Then I'll go ahead and merge.

maxinelasp:
Sure thing, sorry, I wasn't sure if you meant a comment in the code or elsewhere!

stefanv (approved):
@larsoner Sanity check?

stefanv:
> Sure thing, sorry, I wasn't sure if you meant a comment in the code or elsewhere!

Thank you very much. numpydoc can do with a lot of love, so we appreciate your contribution.

maxinelasp:
Any updates on this? Looks like it wasn't merged.

numpydoc/validate.py
```diff
@@ -711,7 +711,12 @@ def validate(obj_name, validator_cls=None, **validator_kwargs):
             errs.append(error("SS03"))
         if doc.summary != doc.summary.lstrip():
             errs.append(error("SS04"))
-        elif doc.is_function_or_method and doc.summary.split(" ")[0][-1] == "s":
+        # Heuristic to check for infinitive verbs - shouldn't end in "s"
+        elif (
+            doc.is_function_or_method
+            and doc.summary.split(" ")[0][-1] == "s"
+            and doc.summary.split(" ")[0][-2] != "s"
```
larsoner, line 718: Just thought of an annoying potential corner case like `"""s blah blah."""` throwing an index error
```suggestion
            doc.is_function_or_method
            and len(doc.summary.split(" ")[0]) > 1
            and doc.summary.split(" ")[0][-1] == "s"
            and doc.summary.split(" ")[0][-2] != "s"
```

larsoner:
Ignore that comment... marking for merge-when-green, thanks in advance @maxinelasp and sorry I missed the ping for review!

larsoner:
FYI I ended up needing to tweak the tests. Nothing too controversial I think -- removed some no-longer-needed ignores, and changed the `Process ...` that was a false alarm before to a `Creates ...` that is a proper hit, and adapted the ignore lines in the tests to ignore it where we ignored `Process ...` before.

PR #622, opened 2025-05-11
Constructor checking for AST validator

mattgebert:
Modifies `validate.py` to include checks for constructors methods (`class.__init__`) implemented by class docstring when using the AST validator, which is essential for hook use (probably the majority use case for numpydoc).  Also fixes nested class references (previously implemented incorrectly).

This pull request now adds checks for the AST class instance, and if using the AST, splits the names of the submodule nested classes, and finds the parent through a AST node `walk` procedure, before allowing the cls_doc check for parameter mismatches.

Solution to #591 and fixed inconsistency in #605.


mattgebert:
I've noted that even despite using `--pre` for pre-release versions of dependencies, we're still not running hook tests in the pre-release jobs. In fact, I can see that the hook tests are not even being collected (though this is working perfectly on my local machine).... I will try adding an explicit call to pytest for hooks.

All pytest unit testing, including `prerelease` and `test` jobs, missing the hook pytest collection ([link](https://github.com/numpy/numpydoc/actions/runs/14956989656/job/42013979718?pr=622) to recent check):
image

Local machine collecting the pytest collection
image
3 extra files: 
 - example_module.py
 - test_utils.py
 - test_validate_hook.py

numpydoc/validate.py
```diff
@@ -639,20 +647,51 @@ def validate(obj_name, validator_cls=None, **validator_kwargs):
...
+                    next_node = module_node
+                    for name in nested_cls_names:
+                        next_node = _find_class_node(next_node, name)
+                    # Get the documentation.
+                    cls_doc = AstValidator(
+                        ast_node=next_node, filename=filename, obj_name=cls_name
+                    )
+                else:
+                    # Ignore edge case: __init__ functions that don't belong to a class.
+                    cls_doc = None
```
stefmolin, line 684: I don't think this is the way to do this. The `DocstringVisitor` class is already traversing the tree and capturing node ancestry (currently, just the name, but after #623, it will be all AST node ancestors). This is then used to instantiate the `AstValidator`, so we should be able to use that information to modify the check for what you are trying to achieve. A lot of the logic here is pretty much doing what those classes are doing, but for a second time, which I would like to avoid.

You may be able to simply filter out any cases that shouldn't be flagged in `DocstringVisitor._get_numpydoc_issues()`, at which point you would have access to the AST node and its ancestry. This is how the ignore rules are applied for the hook.

mattgebert: Brilliant - I was actually hoping this could be the case prior to making these changes. 

Once #623 is merged I'll be happy update this logic and use the ancestry.

rossbar (commented):
@mattgebert can you confirm that you believe this is sufficient to close the regressions reported in #637 and #638? Do we have tests that sufficiently cover the reported failure cases (IYO)?

mattgebert:
@rossbar I believe this PR is ready now, if you want to look over it.

PR #623, opened 2025-05-11
Switch to storing AST nodes on the stack for more accurate method signature check and easy access to parent nodes

stefmolin:
AST nodes have knowledge of their children, but not their parents which is why we use a stack in the `DocstringVisitor`. Currently it is used to track a node's unambiguous name (e.g., `my_module.MyClass.__init__`). This PR changes that to store the AST nodes themselves, which allows us to use a node's ancestry for more robust logic. For example, with this change, we can distinguish between a function and a method, and more precisely determine whether to ignore `self` and `cls` parameters in a signature.

### DIFF
--- a/numpydoc/hooks/validate_docstrings.py
+++ b/numpydoc/hooks/validate_docstrings.py
@@ -33,7 +33,12 @@ class AstValidator(validate.Validator):
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


larsoner (approved):
I'm a little out of my depth here, but it looks reasonable to me, CIs are happy, and it should help with #622 so I'm inclined to merge. @stefanv did you want to look?

PR #626, opened 2025-05-24
Switch to dependency groups

### LINKED ISSUE #625: Thoughts on switching to dependency groups in `pyproject.toml`?
[PEP 735](https://peps.python.org/pep-0735/) added [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups) to `pyproject.toml`, which would remove the need to have multiple requirements files.

This would allow us to hopefully remove those `requirements.txt` files and the logic to generate them. For example, in  `.circleci/config.yml` we could make the following change:

```diff
- python -m pip install --upgrade -r requirements/doc.txt
+ python -m pip install --upgrade --group doc
```

stefmolin:
Closes #625

[PEP 735](https://peps.python.org/pep-0735/) added [dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups) to `pyproject.toml`, which would remove the need to have multiple requirements files. This PR makes that switchover and removes the script that generated those files along with the local pre-commit hook to trigger it.

I also renamed the "developer" group as "dev", which with `uv` at least gets installed automatically by default when building from source. I also included the "doc" and "test" groups inside the "dev" group, so developers have everything bundled.


larsoner:
Thanks @stefmolin !

Issue #497, opened 2023-08-24
Exclude Files in Test Modules from Numpydoc Validation


RNKuhns:
Love the ability to include the `numpydoc` validations in my pre-commit routine.

Sorry I got busy and couldn't help with testing.

As I've been implementing the validations as a pre-commit routine I realized I'm having trouble coming up with a pattern to exclude my test files in the "exclude" option in my pyproject.toml.

I want to include any object in a "tests" directory from validation.

PR #635, opened 2025-06-26
Added "exclude_files" option for pyproject.toml config usage.

mattgebert:
Added an option for a `exclude_files` option in pyproject.toml config that allows `re` regular expression path exclusions.

Unfortunately, most yaml tags use `exclude` to refer to files, which perhaps doesn't apply to toml, but we already use the [`exclude`](https://numpydoc.readthedocs.io/en/latest/validation.html) label to notate object exclusion. I have introduced `exclude_files` instead to achieve the same purpose, which is almost processed identically to `exclude` except being utilized instead when iterating over the `run_hook` filepath list.

I'm not familiar with the [Sphinx build checking ](https://numpydoc.readthedocs.io/en/latest/validation.html#docstring-validation-during-sphinx-build) options that `exclude` currently uses. Do I/we need to implement a mapping for this in that location too?

Solution to Issue #497

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -141,11 +141,11 @@ numpydoc_validation_exclude : set
     ``numpydoc_validation_checks`` is not an empty set.
 numpydoc_validation_exclude_files : set
     A container of strings using :py:mod:`re` syntax specifying path patterns to
-    ignore for docstring validation.
+    ignore for docstring validation, relative to the package root.
     For example, to skip docstring validation for all objects in
     ``tests\``::
 
-        numpydoc_validation_exclude_files = {"$.*tests/.*"}
+        numpydoc_validation_exclude_files = {"^tests/.*$"}
 
     The default is an empty set meaning no paths are excluded from docstring
     validation.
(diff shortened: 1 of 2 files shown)


larsoner:
> I'm not familiar with the Sphinx build checking options that exclude currently uses. Do I/we need to implement a mapping for this in that location too?

I think ideally yes so that the hook and the Sphinx-based checking both function similarly. In principle they should use 99% of the same mechanics so hopefully it won't be too hard to add at the Sphinx end...

numpydoc/numpydoc.py
```diff
@@ -190,6 +192,23 @@ def mangle_docstrings(app, what, name, obj, options, lines):
...
+        # Test the obj to find the module path, and skip the check if it's path is matched by
+        # numpydoc_validation_exclude_files
+        if (
+            app.config.numpydoc_validation_exclude_files
+            and app.config.numpydoc_validation_checks
+        ):
+            excluder = app.config.numpydoc_validation_files_excluder
+            module = inspect.getmodule(obj)
+            try:
+                path = module.__file__ if module else None
```
mattgebert, line 204: Possibly problematic to get global path to file, rather than relative path to a package root? Perhaps this is fine.

mattgebert: @larsoner I think this is concrete - the __file__/__package__/__module__ and package indexing is strongly linked like this. 

I presume doing this
```
>>> import scipy.signal.windows
>>> module = scipy.signal.windows
>>> ".".join(module.__name__.split(".")[1:])
'signal.windows'
```
is more risky, but I can't foresee a case where a module directory path isn't directly linked the module name?

Issue #637, opened 2025-06-27
Exception using dataclasses with v1.9


EthanAtLL:
Running `uv run doctest.py` on

```py
# /// script
# dependencies = [
#   "numpydoc==1.8.0",
# ]
# ///
from dataclasses import dataclass
from numpydoc.validate import validate, get_doc_object

@dataclass
class MyClass:
    """Docs.
    """
    x: int

if __name__=='__main__':
    doc_obj = get_doc_object(MyClass)
    x = get_doc_object(MyClass.__init__)
    print(validate(x))
```

works yielding `{'type': 'function', 'docstring': '', 'deprecated': False, 'file': '<string>', 'file_line': None, 'errors': [('GL08', 'The object does not have a docstring')], 'examples_errors': ''}` but changing to numpydoc==1.90 gives

```
Traceback (most recent call last):
  File ".../doctest.py", line 18, in <module>
    print(validate(x))
  File ".../uv/environments-v2/doctest-76a1dd0b6c7db47d/lib/python3.9/site-packages/numpydoc/validate.py", line 648, in validate
    cls = Validator._load_obj(f"{doc.code_obj.__module__}.{cls_name}")
  File ".../uv/environments-v2/doctest-76a1dd0b6c7db47d/lib/python3.9/site-packages/numpydoc/validate.py", line 268, in _load_obj
    obj = getattr(obj, part)
AttributeError: module '__main__' has no attribute '__create_fn__'
```

The error only arises with the `@dataclass` annotation.

DerWeh:
I have the same issue, in case further logs help you:
https://app.readthedocs.org/projects/gftools/builds/28675578/
https://github.com/DerWeh/gftools/actions/runs/15946217959/job/44980605108?pr=45

eaubin:
@mattgebert Thank you for looking into this! @larsoner With 3.9 EOLing soon, requiring 3.10+ makes sense. Feel free to close.

EthanAtLL:
@mattgebert Thank you for looking into this! @larsoner With 3.9 EOLing soon, requiring 3.10+ makes sense.

PR #641, opened 2025-07-27
build(pyproject.toml): drop support for python 3.9 (require 3.10+)

mattgebert:
### Purpose:
In upgrading via #592 and #622, the new functionality for checking constructor docstrings fulfilled by the class docstring requires stable use of `__qualname__` for the Validator class. 

### The problem:
This property is not supported in dataclasses in python3.9, but from 3.10 onward (solves #637).

### Suggestion:
Drop support for Python3.9, which [EOL is Oct 2025](https://devguide.python.org/versions/), so #622 can be validated and merged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu]
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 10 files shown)


larsoner:
@mattgebert now that we require 3.10 the style checkers want you to use `|` for some type checks, can you do it here? Otherwise I can push to do it

larsoner:
Thanks for the ping @mattgebert . I don't get notified about commits so had no idea you'd made the changes!

PR #654, opened 2025-10-19
Remove duplicate dependency in pyproject.toml

stefmolin:
Remove `tomli` dev dependency for Python versions >= 3.11 since it is already in the package dependencies section (line 31)


stefanv:
Thanks!

PR #669, opened 2025-12-02
pyproject: include example_module.py

### LINKED ISSUE #668: No such file or directory: 'numpydoc/tests/hooks/example_module.py'
running tests fomr the pypi tarball fails because example_module.py is not included in the archive:
https://github.com/numpy/numpydoc/blob/main/pyproject.toml#L171

xref https://github.com/conda-forge/numpydoc-feedstock/pull/34

jschueller:
Closes #668

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,7 +168,7 @@ attr = 'numpydoc.__version__'
 [tool.setuptools.package-data]
 numpydoc = [
     'tests/test_*.py',
-    'tests/hooks/test_*.py',
+    'tests/hooks/*.py',
     'tests/tinybuild/Makefile',
     'tests/tinybuild/index.rst',
     'tests/tinybuild/*.py',


rossbar:
Unfortunately I think the fix may be more involved - I cherry-picked this over to #670 to test that it fixes the issue, but the test is still failing. I *believe* the issue is the test design itself - AFAICT [the example_module fixture](https://github.com/numpy/numpydoc/blob/16a20aad58d5532fb12c2a87af0eb4ea290bacb4/numpydoc/tests/hooks/test_validate_hook.py#L13) implicitly expects the tests are run from the source directory rather than pulling paths relative to the install location. This should be fixable with `importlib.resources`, but I'm having trouble groking the original test intent with the relative paths.  @stefmolin @mattgebert any insight?

stefmolin:
> Unfortunately I think the fix may be more involved - I cherry-picked this over to #670 to test that it fixes the issue, but the test is still failing. I _believe_ the issue is the test design itself - AFAICT [the example_module fixture](https://github.com/numpy/numpydoc/blob/16a20aad58d5532fb12c2a87af0eb4ea290bacb4/numpydoc/tests/hooks/test_validate_hook.py#L13) implicitly expects the tests are run from the source directory rather than pulling paths relative to the install location. This should be fixable with `importlib.resources`, but I'm having trouble groking the original test intent with the relative paths. @stefmolin @mattgebert any insight?

It was relative paths so the expected output would be matchable regardless of who ran the tests since it prints the full path. Now that we use the f-string we shouldn't need to keep them relative.

rossbar:
I cherry-picked this and included it in #670 so I'll close this. Thanks for the fix @jschueller !

PR #670, opened 2025-12-02
TST,CI: Add workflow to check sdist.

rossbar:
Expect this initial commit to fail - verifying that the tests work as expected.

### DIFF
--- a/numpydoc/tests/hooks/test_validate_hook.py
+++ b/numpydoc/tests/hooks/test_validate_hook.py
@@ -13,6 +13,10 @@ from numpydoc.hooks.validate_docstrings import run_hook
 
 @pytest.fixture
 def example_module(request):
+    # TODO: When Python3.13 is the minimum version supported version, this
+    # can be simplified to:
+    # with importlib.resources.path(numpydoc, "tests/hooks/example_module.py") as fpath:
+    #     fullpath = str(fpath)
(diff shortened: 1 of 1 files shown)


rossbar:
This recipe behaves as expected (i.e. fails) locally, but something wonky is going on with the actions (not sure why they're not running).

Anyways - I just noticed that there already exists a `sdist_wheel` job - perhaps it makes more sense to add a testing step to that job instead? WDYT @larsoner ?

larsoner:
> Anyways - I just noticed that there already exists a sdist_wheel job - perhaps it makes more sense to add a testing step to that job instead? WDYT @larsoner ?

Sure!

> This recipe behaves as expected (i.e. fails) locally, but something wonky is going on with the actions (not sure why they're not running).

In the actions tab you can see it's unhappy about some syntax

https://github.com/numpy/numpydoc/actions/runs/19871137814

```
(Line: 214, Col: 9): There's not enough info to determine what you meant. Add one of these properties: run, shell, uses, with, working-directory
```

rossbar:
A final thought: ultimately I think switching to `cibuildwheel` would significantly simplify both the testing and release process for wheels/source dists. That touches more infra though... this "hotfix" should at least be enough to provide temporary assurance that numpydoc isn't missing anything in the package.

larsoner (approved):

Issue #511, opened 2023-10-20
"Handler <function mangle_docstrings at 0x7f64b5ba57b8> for event 'autodoc-process-docstring' threw an exception" again


phxnsharp:
The recent release of numpydoc 1.6.0 caused our builds to start failing with the exact error from #268. Pegging numpydoc to 1.5.0 allows our builds to continue. 

Excerpt from pyproject.toml:

```toml
[tool.poetry.dependencies]
python = ">=3.8,<4"
pydantic = "^2.3.0"
anyio = ">=3.7.1,<5.0.0"


[tool.poetry.group.doc]
optional = true
[tool.poetry.group.doc.dependencies]
ansys-sphinx-theme = "^0.9.8"
autodoc_pydantic = "^2.0"
httpx = "^0.23.3"
numpydoc = "^1.4.0"
sphinx = "^6.2.1"
sphinx-autobuild = "^2021.3.14"
sphinx-autodoc-typehints = "^1.23.0, !=1.23.4"
sphinx-copybutton = "^0.5.2"
sphinx_design = "^0.4.1"
sphinx_code_tabs = "^0.5.3"
sphinx-gallery = "^0.13.0"
sphinx_mdinclude= "^0.5.3"
sphinx-notfound-page = "^0.8.3"
sphinx-tabs = ">=1.2.1,<3.5.0"
sphinxcontrib-websupport = "^1.2.4"
sphinxemoji = "^0.2.0"
toml = "^0.10.2" # Needed by conf.py
# `sphinx-toolbox` requires `cachecontrol`, which is incompatible with `urllib3>=2`
# See https://github.com/ionrock/cachecontrol/issues/292.
urllib3 = "<2"
tox = "^4.4.6"
```

Sphinx v6.2.1

Error:

```
[app] emitting event: 'build-finished'(ExtensionError("Handler <function mangle_docstrings at 0x7f512e29b760> for event 'autodoc-process-d

Traceback (most recent call last):
  File "/home/nsharp/ver/bdm-python-api/.venv/lib/python3.10/site-packages/sphinx/events.py", line 96, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/nsharp/ver/bdm-python-api/.venv/lib/python3.10/site-packages/numpydoc/numpydoc.py", line 214, in mangle_docstrings
    report = validate(doc)
[...]

PR #675, opened 2026-01-12
fix: resolve numpydoc StopIteration error for async functions

LinyanZhang:


LinyanZhang:
Sorry for the late response! Yes, this fixes #511. I've added a test case (test_async) to DecoratorClass in test_validate.py that fails on main but passes on this PR.

larsoner (approved):
Looks good, marking for merge-when-green, thanks in advance @LinyanZhang !

PR #688, opened 2026-05-21
fix: handle Sphinx 9.x new autodoc options key format and sentinels

### LINKED ISSUE #671: compatibility with new autodoc implementation ?
sphinx 9.x has rewritten its autodoc extension,
is there anything left to do in numpydoc to leverage the new api ?

I noticed when upgrading to sphinx 9.x as I had to switch autodoc_use_legacy_class_based=True to use the legacy class-based API:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_use_legacy_class_based

else it gave me lots of warnings about references not found: [ref.obj] for stuff that should be filtered by `:exclude-members:` (nitpicky mode), eg:
```
/tmp/build/python/src/site-packages/persalys/persalys.py:docstring of persalys.persalys.Analysis:8: WARNING: py:obj reference target not found: thisown [ref.obj]
```

so is this a known task to handle the new default, or should I try to write a minimal reproducer ?

jschueller:
Sphinx 9.x's new non-legacy autodoc uses underscored keys (e.g., `exclude_members`) and the `EMPTY` sentinel instead of hyphenated keys (`exclude-members`) and the `ALL` sentinel used by older Sphinx. This caused `:exclude-members:` to be silently ignored by numpydoc, producing spurious `py:obj reference target not found` warnings for excluded inherited members in nitpicky mode.

- Update `ClassDoc.__init__` to check both `exclude_members` and `exclude-members` keys, and normalize `EMPTY` to `ALL`
- Update `mangle_docstrings` to use `vars()` instead of the deprecated Mapping interface for `_AutoDocumenterOptions`

Closes #671

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -628,8 +628,17 @@ class ClassDoc(NumpyDocString):
 
         if "sphinx" in sys.modules:
             from sphinx.ext.autodoc import ALL
+
+            try:
+                from sphinx.ext.autodoc._sentinels import EMPTY
+            except ImportError:
+                try:
+                    from sphinx.ext.autodoc import EMPTY
+                except ImportError:
+                    EMPTY = object()
         else:
             ALL = object()
+            EMPTY = object()
 
         if config is None:
             config = {}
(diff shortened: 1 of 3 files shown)


larsoner:
Looks reasonable to me!

See any easy way to add a test? Maybe by triggering the `WARNING` log line from #671 and with `capsys` make sure `WARNING` isn't in the output anymore (but was in `main`)?

jschueller:
done, we can test mangle_docstrings directly

rossbar (approved):
LGTM too, thanks for digging into this @jschueller !

PR #691, opened 2026-05-27
Add support for custom section names

brockdyer03:
Motivated by #202.

## WIP Implementation

The implementation I've gotten to work for now is quite rough, but basically can be boiled down to two steps for users. The first is to provide a custom template file, which follows the form of [`numpydoc_docstring.rst`](https://github.com/numpy/numpydoc/blob/81268dc3b160e43c6a6e007dc6ead9385c83e19c/numpydoc/templates/numpydoc_docstring.rst), and just requires that the user add in whichever new section names they want. This template file is then picked up by the variable in `conf.py` called `numpydoc_template_file`.

The other part of this is adding in the desired formatting of the new section. This is done through setting `numpydoc_extra_sections` in `conf.py`, which is a dictionary with the keys being the section names and the values being the desired format. For example, you could have
```python
numpydoc_extra_sections = {
    "Note": "notes",
    "Members": "member_list",
}
```

These format values map to the existing string functions in `docscrape_sphinx.SphinxDocString`, and are matched like so:

```python
"param_list"  : self._str_param_list(section)
"member_list" : self._str_member_list(section)
"returns"     : self._str_returns(section)
"warnings"    : self._str_warnings(section)
"see_also"    : self._str_see_also(section)
"notes"       : self._str_section(section)
"references"  : self._str_references(section)
"examples"    : self._str_examples(section)
```

## To Do

I am not a huge fan of users needing to supply two separate values to get a new section in, however I haven't figured out a way to add in extra sections that I like.

Another thing to do is to update the validation code to recognize custom sections, if that is desired.

## Example

I have a minimal example [in a branch in my own repository](https://github.com/brockdyer03/quadrupole/tree/docs), where all I've done is very simply change `Notes` to `Note` in the `Element` enum.

### DIFF
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -11,6 +11,9 @@ from collections.abc import Callable, Mapping
 from functools import cached_property
 from warnings import warn
 
+from sphinx.util import logging
+
+logger = logging.getLogger(__name__)
 
 def strip_blank_lines(l):
     "Remove leading and trailing blank lines from a list of lines"
@@ -139,6 +142,18 @@ class NumpyDocString(Mapping):
     def __init__(self, docstring, config=None):
         orig_docstring = docstring
         docstring = textwrap.dedent(docstring).split("\n")
+
+        if config is not None:
+            extra_sections = config.get("extra_sections", dict())
+            for section, fmt in extra_sections.items():
+                if fmt in ("param_list", "member_list", "returns", "warnings", "see_also", "notes"):
+                    default = []
+                elif fmt in ("references", "examples"):
+                    default = ""
+                else:
+                    logger.warning("Unrecognized section format %r for section %s", fmt, section)
+
+                NumpyDocString.sections.update({section: default})
 
         self._doc = Reader(docstring)
         self._parsed_data = copy.deepcopy(self.sections)
--- a/numpydoc/docscrape.py
+++ b/numpydoc/docscrape.py
@@ -15,6 +15,7 @@ from sphinx.util import logging
 
 logger = logging.getLogger(__name__)
 
+
 def strip_blank_lines(l):
     "Remove leading and trailing blank lines from a list of lines"
     while l and not l[0].strip():
@@ -146,12 +147,21 @@ class NumpyDocString(Mapping):
         if config is not None:
             extra_sections = config.get("extra_sections", dict())
             for section, fmt in extra_sections.items():
-                if fmt in ("param_list", "member_list", "returns", "warnings", "see_also", "notes"):
+                if fmt in (
+                    "param_list",
+                    "member_list",
+                    "returns",
+                    "warnings",
+                    "see_also",
+                    "notes",
+                ):
                     default = []
                 elif fmt in ("references", "examples"):
                     default = ""
                 else:
-                    logger.warning("Unrecognized section format %r for section %s", fmt, section)
+                    logger.warning(
+                        "Unrecognized section format %r for section %s", fmt, section
+                    )
 
                 NumpyDocString.sections.update({section: default})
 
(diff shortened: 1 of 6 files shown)


numpydoc/docscrape.py
```diff
@@ -11,6 +11,9 @@
 from functools import cached_property
 from warnings import warn
 
+from sphinx.util import logging
```
stefanv, line 14: We're trying to move away from having a Sphinx dependency, so probably don't want this.

numpydoc/docscrape.py
stefanv, line 143: We need to ensure that `config` is documented somewhere.

numpydoc/docscrape.py
```diff
@@ -140,6 +143,18 @@ def __init__(self, docstring, config=None):
         orig_docstring = docstring
         docstring = textwrap.dedent(docstring).split("\n")
 
+        if config is not None:
+            extra_sections = config.get("extra_sections", dict())
+            for section, fmt in extra_sections.items():
```
stefanv, line 149: Can probably handle this with set intersections.

numpydoc/docscrape_sphinx.py
```diff
@@ -29,12 +33,20 @@ def load_config(self, config):
         self.xref_param_type = config.get("xref_param_type", False)
         self.xref_aliases = config.get("xref_aliases", dict())
         self.xref_ignore = config.get("xref_ignore", set())
-        self.template = config.get("template", None)
+        self.extra_sections = config.get("extra_sections", dict())
+        self.template = config.get("template")
+        self.template_file = config.get("template_file")
```
stefanv, line 37: Are these changes related to the PR (looks like they are)? If so, document along with other config options.

numpydoc/docscrape.py
brockdyer03: I'd be fine removing it if there's a replacement that has already been agreed on.

numpydoc/docscrape_sphinx.py
```diff
@@ -377,6 +389,31 @@ def __str__(self, indent=0, func_role="obj"):
             "references": self._str_references(),
             "examples": self._str_examples(),
         }
+
+        for section, fmt in self.extra_sections.items():
+            if not self.get(section):
```
stefanv, line 393: This change doesn't look quite right; ignores the dictionary above?

brockdyer03: The dictionary above it is the standard numpydoc sections, which I suppose can be hard-coded for now. After those are set, I iterate over the *extra* sections that the user provides, and use them to update the above dictionary.

stefanv: OK, but do we need to re-handle existing sections in there, or can those be removed and handled via the standard mechanism?

brockdyer03: The existing sections aren't touched there. The reason it looks like it is because I am re-using the names of the `_str_<section_format>()` functions to select the format, which are also the identifiers in `numpydoc_docstring.rst`.

numpydoc/docscrape.py
brockdyer03: I agree. I am not 100% sure what the type of `config` is everywhere, but I believe it's just a simple `dict`. I am more sure that it is at least supposed to support the `MutableMapping` ABC.

numpydoc/docscrape_sphinx.py
brockdyer03: Yes. I originally had put the code for the extra templating information there, but realized that it is additionally set in `get_doc_object()` at the end of that module. I am unsure if it *needs* to be in both places, but I think it is probably good to have it in both since the one you commented on sets the attributes of `SphinxDocString`, which already has attributes for the other config options.

brockdyer03:
@stefanv, just pinging to check if this PR is still being considered?

PR #706, opened 2026-09-07
FIX: resolve attribute cross-links in the class scope first

lazerg:
On a class page numpydoc renders every entry of the Attributes list as ``:obj:`name <name>` ``. Sphinx resolves that
target against the current module before the current class, so an attribute whose name is also a module-level object
links to the wrong page. On the `numpy.ufunc` page the `identity` attribute links to the `numpy.identity` function
instead of `numpy.ufunc.identity`, and it has been doing that since at least numpy 1.20.

Prefixing the target with a dot makes the reference specific, so Sphinx tries the class scope first and falls back to
the module scope exactly as before when the attribute is not documented itself.

Reported in numpy/numpy#32532.


larsoner:
Seems totally reasonable (and I had no idea about this `"."` behavior!). Would it be possible to add some test that fails on `main` but passes on this PR?

lazerg:
I added test_attribute_link_is_class_scoped in numpydoc/tests/test_docscrape.py. It builds a class with an attribute and checks the rendered link uses the dot prefix. I ran it against main: it fails (link comes out as `<identity>`). I ran it against this branch: it passes. Commit 8e7513c.

larsoner (approved):

larsoner:
Thanks @lazerg !

PR #707, opened 2026-09-15
Prep for release

larsoner:
1. Use proper dynamic versioning to avoid need for bumping version after every release (including `git-archival` stuff so installing from zipball still works)
2. Use `changelist` to generate changelist
3. Remove bot PRs from changelist


larsoner:
Verified locally that scipy numpy networkx matplotlib scikit-image, so I'll merge and cut the release!

PR #708, opened 2026-09-15
Run integration tests in CircleCI

larsoner:
I didn't like having to run the integration tests manually before cutting a release, so I asked Claude Opus 5 to try adding an integration test to CircleCI. After some iteration this is where things ended up -- a 300-line helper script plus a small CircleCI config change to enable networkx, scikit-image, matplotlib, and numpy builds on PRs. The script needs to be that long to deal with running doc builds with `-W` while doing stuff like skipping sphinx-gallery (matplotlib) etc. I looked through the changes and think they're reasonable, so I'm opening to see if it works :crossed_fingers: 

Not 100% sure this is the best approach but it seems maybe worthwhile if it runs quickly enough. We'll see how fast CircleCI builds these :rocket: (the new `.gen3` runners have been quite fast in my experience so far)


larsoner:
I made it so the tests only run if `integration-tests` label is added, or it's a `main` build. Not 100% sure it's necessary since mpl builds in 3m but I also suspect regressions caught by these CIs will be fairly rare so I think it's an okay compromise to save some CI cycles.

rossbar (approved):
Thanks @larsoner - I'm in favor of trying this out; it's nice to have this scripted up and runnable in CI rather than requiring release managers/devs to do so manually!

I tried things locally and it went okay, more or less. The script is loaded with piles of abstraction/indirection that claude is addicted to, but this isn't user-facing at all; I'm not to worried about the readability/maintainability of dev tooling.

I say let's give it a try!

